PR #293, opened 2024-11-06
initial commit of flow track code with STEREO-COR2 inputs

raphael-attie:
## PR summary

Flow Track code with STEREO-COR2-tested inputs, modifiable for PUNCH. 

## Todos

- Adapt inputs with PUNCH with polar-remapped synthetic data and header
- Unit tests


jmbhughes:
This looks great! We'll need to configure a few things to make it run in the automated pipeline. I'm happy to take care of those things for you once you're happy with the state of the code. Thanks! 

Also, the CI is only failing because of some Codecov submission. We'll need to add tests, but this addition didn't break any of our existing tests.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+    sparsity = 2  # Frame skip interval for averaging
+    n_ofs = 151  # Number of spatial offsets for cross-correlation
+    dr = 2  # Pixel offset increment per sample
+    of0 = dt * 9  # Central offset based on expected wind speed (300 km/s)
+    expected_kps_windspeed = 300  # Expected wind speed in km/s
+    chw = 0.5  # Half-width of each radial band in solar radii
+    nlatbins = 36  # Number of angular bins (latitude bins)
+    # Define radial band centers in solar radii
+    ycens = np.array([7, 7.5, 8, 8.5, 9, 9.5, 10, 10.5, 11, 11.5, 12, 12.5, 13, 13.5, 14])
+    rbands = [0, 4, 8, 14]  # Indices of radial bands to visualize
```
jmbhughes, line 403: I'd like to understand these parameters a bit better. I also lean towards the side of longer, more explicit variable names so it's clear what they mean when they're used. For example, `chw` might be changed to `radial_half_width`. Those are pretty straight forward refactorings once we have everything working right.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+
+    Returns
+    -------
+    list
+        Lower and upper Numpy array indices of the band.
+    """
+    y0_image_px = 1
+    y0_cen_px = header['CRPIX2']
+    y0_cen_rs = header['CRVAL2']
+    px_scale = header['CDELT2']
```
jmbhughes, line 77: Is this assuming `CDELT2` is in R_sun/pixel?

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+    np.ndarray
+        2D image array from the specified FITS file.
+    """
+    with fits.open(filepath) as hdul:
+        hdul.verify('fix')
+        data = hdul[hdu].data
+    return data
+
+
+def read_header(filepath: str, hdu: int = 1) -> fits.header.Header:
```
jmbhughes, line 32: I might suggest reading the data and the header at the same time instead of opening the file twice. Very minor optimization.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+        Index of the header data unit storing the data array, by default 1.
+
+    Returns
+    -------
+    np.ndarray
+        2D image array from the specified FITS file.
+    """
+    with fits.open(filepath) as hdul:
+        hdul.verify('fix')
+        data = hdul[hdu].data
```
jmbhughes, line 28: Note that our data sometimes come in cubes at levels 2 and 3. So you'll have an array with dimensions (polarization, spatial, spatial). The data I shared was (3, 4096 ,4096). Level 3 polarized files are actually (2, 4096, 4096) because they're in B/pB instead of MZP polarization. Long story short, I added a `layer` parameter to the `read_fits` so that I could get the specific image out with `data = hdul[hdu].data[layer]`.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+    np.ndarray
+        Accumulated cross-correlation array over all offsets.
+    """
+    acc = np.zeros((len(offsets), a.shape[0], a.shape[1]), dtype=float)
+    for jj, offset_index in enumerate(offsets):
+        this_of = int(dr * (offset_index - (len(offsets) - 1) / 2)) + of0
+        ofa = int(this_of / 2)
+        ofb = int(this_of) - ofa
+
+        if ofa < 0:
```
jmbhughes, line 136: Could you add a comment in the code on where these pad shapes come from? It'll be helpful for when we return to the code later.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+    Returns
+    -------
+    tuple
+        Tuple containing:
+        - np.ndarray : Average speed per angular bin for each radial band.
+        - np.ndarray : Sigma (standard deviation) of speed per angular bin for each radial band.
+    """
+    ylohi = calc_ylims(ycen_band_rs, chw, header)
+    # Determine spike location (index of the correlation peak) in the cross-correlation array
+    spike_location = np.where(x_kps < 0)[0].max() + 2
```
jmbhughes, line 268: Where does the +2 come from? I assume that'll never cause an out of bounds error for indexing by `spike_location`?

scripts/flow_track_correl.py
jmbhughes: Also, what if we have an image missing because we couldn't downlink from the satellite or something? This seems to assume the images are regularly spaced. Would it be helpful to use the observation time from the header?

raphael-attie: These parameter list are COR2-related and the intent is that once I have the mosaic data with the full extent of the field of view, I would like to reduce the numbers of parameters that we need to input. In any case, I can change their names for longer names, I also prefer more explicit, longer names for the sake of quickly understanding what's going on without having to look at the documentation.

scripts/flow_track_correl.py
raphael-attie: Yes. Need adaptation if mosaic header has different units.

scripts/flow_track_correl.py
raphael-attie: Had that in mind too. I'll refactor.

scripts/flow_track_correl.py
raphael-attie: Noted. That's the kind of information I was hoping to get updated on.

scripts/flow_track_correl.py
raphael-attie: Will do. That was a rather "arbitrary" decision from Craig's original code to choose in which direction to shift the images with respect to each other (the first value is... 42... lol), but depending on the dimensions of the polar-remapped data of the mosaic data, I may adapt that for both clarity and efficiency.

scripts/flow_track_correl.py
```diff
@@ -0,0 +1,414 @@
...
+
+    avg_speeds = []
+    sigmas = []
+    for kk, (ylo, yhi) in enumerate(zip(*ylohi)):
+        acc_k = acc[:, int(ylo):int(yhi) + 1, ...].mean(axis=1)
+        latbinsize = acc_k.shape[1] // nlatbins
+        avcor_rbins_theta = acc_k.reshape(acc_k.shape[0], latbinsize, nlatbins)
```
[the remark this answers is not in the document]
raphael-attie: clarifying: ntlabins was chosen to 36 so I end up with 100 points per angular bin **[for STEREO/COR2]**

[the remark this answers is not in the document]
jmbhughes: That makes sense. The WCS for mosaics is defined by:

```
SIMPLE  =                    T / conforms to FITS standard                     
BITPIX  =                    8 / array data type                               
NAXIS   =                    0 / number of array dimensions                    
NAXIS1  =                 4096 / length of the first axis                      
NAXIS2  =                 4096 / length of the second axis                    
WCSAXES =                    2 / Number of coordinate axes                     
CRPIX1  =               2048.0 / Pixel coordinate of reference point          
CRPIX2  =               2048.0 / Pixel coordinate of reference point           
CDELT1  =               0.0225 / [deg] Coordinate increment at reference point 
CDELT2  =               0.0225 / [deg] Coordinate increment at reference point 
CUNIT1  = 'deg'                / Units of coordinate increment and value       
CUNIT2  = 'deg'                / Units of coordinate increment and value       
CTYPE1  = 'HPLN-ARC'           / Coordinate type zenithal/azimuthal equidistant 
CTYPE2  = 'HPLT-ARC'           / Coordinate type zenithal/azimuthal equidistant 
CRVAL1  =                  0.0 / [deg] Coordinate value at reference point     
CRVAL2  =                  0.0 / [deg] Coordinate value at reference point     
LONPOLE =                180.0 / [deg] Native longitude of celestial pole      
LATPOLE =                  0.0 / [deg] Native latitude of celestial pole       
MJDREF  =                  0.0 / [d] MJD of fiducial time                      
END                            
```

Note that a third FITS dimension is appended for polarized data. Other than that, the WCS we reproject to is fixed. Note that our mosaics are not polar maps. We can talk about that later.

scripts/flow_track_correl.py
jmbhughes: 42. The answer!

scripts/flow_track_correl.py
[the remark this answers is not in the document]
raphael-attie: I will have to modify the coordinate conversion in calc_ylims().

scripts/flow_track_correl.py
raphael-attie: Will add comment on the padding too, not just the values of ofa and ofb

raphael-attie:
@lowderchris 

Can you document further what's going on with this code block from your polar_projections_punch notebook? 

```python
# Calculate the number of bins in radius and azimuth
radius_bin_size = map_punch.scale[0].to_value('arcsec/pix')
max_radius_arcsec = 45 * 60 * 60
num_radius_bins = int(max_radius_arcsec / radius_bin_size)
num_azimuth_bins = 720 * 3
```

In particular, the hardcoded numbers, what do they correspond to?

lowderchris:
@raphael-attie Absolutely! Sorry for the hardcoded numbers there. I should have been using astropy units more thoroughly to make things clear.

- The maximum radius is 45 degrees to arcsecs
- The azimuthal bins are just making that a multiple of 720 for when I was testing that

For the polar header,

- crpix1 is setting that at the center of the center pixel, hence the division by 2 and addition of half a pixel
- crpix2 is doing something similar, but using the specification of the number of bins, and the bin size (90 degree offset in azimuth, 3600 arcsec / degree, etc)
- crval1 is straightforward
- crval2 is taking that offset in azimuth by 90 degrees, and converting that to arcsec

Hope that helps!

raphael-attie:
> * crval2 is taking that offset in azimuth by 90 degrees, and converting that to arcsec

@lowderchris I'm a little confused on that last item. Why is crval2 about azimuth? Isn't crval2 about the radial axis, not azimuthal?

lowderchris:
> > * crval2 is taking that offset in azimuth by 90 degrees, and converting that to arcsec
> 
> @lowderchris I'm a little confused on that last item. Why is crval2 about azimuth? Isn't crval2 about the radial axis, not azimuthal?

Ah yes, sorry, flipped those two in my mind for a second. Yeah, the first axis parameters are azimuthal, and the second axis is radial.

jmbhughes:
Thanks, Raphael! We have incorporated this code. We may have more questions as we continue testing.

PR #320, opened 2024-11-22
When generating WCSes, set NAXIS/array_shape/etc.

svank:
## PR summary

Ran into this when testing a `remove_starfield` speedup---generated WCSes don't have their NAXISN keywords set, and therefore don't have `pixel_shape` or `array_shape`.  On the `remove_starfield` side that can and should and now is handled. But this seems like a might-as-well-have on the `punchbowl` side.


jmbhughes (approved):

jmbhughes:
There are more spots in punchbowl this matters, so I'll open an issue.

PR #323, opened 2024-11-23
Reproject starfield sample and uncertainty in one fell swoop

svank:
## PR summary

Handling uncertainty in the starfield subtraction was just subtracting the actual starfield from the uncertainty through a second call to `starfield_model.subtract_from_image`. Since it's still the same WCS and transformation, we might as well do it in the same call as for the actual starfield subtraction. That takes things from 10 min to 7m45s.

This does not address the question of what's correct for the uncertainty at this step. It was subtracting the actual starfield, and this PR takes the liberty of changing it to subtract the starfield's uncertainty. Still not correct, but maybe closer.

So the uncertainty comes out different, but the data layer comes out bit-for-bit identical.

### DIFF
--- a/punchbowl/level3/stellar.py
+++ b/punchbowl/level3/stellar.py
@@ -144,22 +144,17 @@ def subtract_starfield_background_task(data_object: NDCube,
         data_wcs = calculate_celestial_wcs_from_helio(data_object.wcs.celestial,
                                                       data_object.meta.astropy_time,
                                                       data_object.data.shape[-2:])
-        starfield_model = Starfield(star_datacube.data, star_datacube.wcs.celestial)
+        starfield_model = Starfield(np.stack((star_datacube.data, star_datacube.uncertainty.array)),
+                                    star_datacube.wcs.celestial)
 
-        starfield_subtracted_data = starfield_model.subtract_from_image(
-            NDCube(data=data_object.data,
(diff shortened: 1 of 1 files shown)


jmbhughes:
Progress! Does using this require a new version of remove_starfield? If so, we should update our pinned version in the server requirements.

svank:
I also have pending changes in remove_starfield that will gain another minute or two, and thoughts on next steps for Monday.

svank:
I wanted to check the `remove_starfield` changes a little bit more, but if you just want to crank stuff out this weekend you can pull the `main` branch. The "blur the input data" stage just directly convolves the data with the right Gaussian, instead of reprojecting it into its own frame twice. As long as there aren't `nan`s or `inf`s in the data, it uses FFT convolution for a big speed gain.

jmbhughes (approved):

jmbhughes:
> As long as there aren't `nan`s or `inf`s in the data, it uses FFT convolution for a big speed gain (on that step).

We'll want to make sure this is true in punchbowl. Uncertainty can have `inf` to indicate complete uncertainty, but we can just make the uncertainty a super big number if it speeds things up.

svank:
We could also swap out uncertainty `inf`s in this stage if it's useful to have them earlier in the pipeline.

PR #324, opened 2024-11-25
Unpack uncertainties faster

svank:
## PR summary

Before I realized that most of the starfield read-in time was decompression, I tried this. This saves 1.3 seconds for loading the starfield map (from 2 s to .7 s) by avoid memory allocations, but maybe saving a smidge of time for every FITS read-in will add up over the whole pipeline.


svank:
This exposed a bug in one of my tests, actually---I was lopping off a dimension but it was broadcasting back and working out.

jmbhughes (approved):

PR #325, opened 2024-11-25
Turn velocity plot test into a figure test

svank:
## PR summary

It's a figure test now


svank:
Note that pytest needs to have '--mpl' added to the command line to actually run the comparison. I like to make that happen by default like [this](https://github.com/svank/wispr_analysis/blob/main/pytest.ini)

jmbhughes:
@svank can you look at the CI for this? Thanks!

svank:
Looks like this repo doesn't use `pytest-arraydiff`, so the `--arraydiff` flag was being rejected. Let's try this.

jmbhughes (approved):

PR #333, opened 2024-11-27
Fixes merging, improves f corona modeling

jmbhughes:


jmbhughes:
pre-commit.ci autofix

PR #338, opened 2024-12-04
Change despike settings

svank:
## PR summary

Is this how you change the despike settings that Prefect uses?

I ran `/d0/punchsoc/gamera_data/0/PM2/2024/12/29/PUNCH_L0_PM2_20241229023210_v1.fits` through despiking with these settings and it's catching like the center pixel of the brightest stars but otherwise it seems to be flagging streaks but not stars.

I need to leave now so I can't check it with any `CR`, `PP`, or `PZ` images, but this is here if you want to cross your fingers and run it overnight.


svank:
OK, this has been tested with `PZ`, `PM` and `PP`, and might be fine for `CR` if not for PSF artifacts.

jmbhughes (approved):

PR #360, opened 2025-02-09
Update binder.yaml

jmbhughes:
This makes binder work maybe... just copies punchbowl into the source directory so it's available. That *might* work.


jmbhughes:
pre-commit.ci autofix

PR #375, opened 2025-03-04
Removes STEAM data products

jmbhughes:


lowderchris (approved):
Gone but not forgotten.

PR #383, opened 2025-03-09
Fixes ruff noqa statement

jmbhughes:
There was an extra statement that cased a Ruff failure. This fixes that.


jmbhughes:
pre-commit.ci autofix

jmbhughes:
@punch-mission/swri-soc Something funny is happening where it passes and fails kind of at the same time. Maybe it's being checked with different ruff versions.

jmbhughes:
pre-commit.ci autofix

lowderchris (approved):

PR #384, opened 2025-03-10
Make cmap a variable

jmbhughes:
Make the colormap a variable instead of just a function. 

@lowderchris we should make sure documents are updated accordingly.


jmbhughes:
pre-commit.ci autofix

lowderchris (approved):

PR #389, opened 2025-03-14
Create an aurora mask product

jmbhughes:


jmbhughes:
Will resolve in a future PR

PR #390, opened 2025-03-18
Fix radec wcs

svank:
## PR summary

Here we go!

## Todos

Two quirks to think about at some point

- [ ] The celestial WCS says it's ICRS, even though it's really GCRS
- [ ] The celestial WCS can have the observation date set, but it doesn't pass through to SkyCoords from `pixel_to_world` since ICRS doesn't care about your observation date

## Test plan

Run the tests!

## Breaking changes

None, but I think we should stop returning the P angle from `calculate_helio_wcs_from_celestial`

### DIFF
--- a/punchbowl/data/tests/test_wcs.py
+++ b/punchbowl/data/tests/test_wcs.py
@@ -208,67 +208,6 @@ def test_load_trefoil_wcs():
     assert isinstance(trefoil_wcs, WCS)
 
 
-def test_helio_celestial_wcs():
-    # header = fits.Header.fromtextfile(os.path.join(_ROOT, "example_header.txt"))
-    #
-    # date_obs = Time(header['DATE-OBS'])
-    date_obs = Time("2021-06-20T00:00:00.000", format='isot', scale='utc')
-    #
-    wcs_helio = WCS({"CRVAL1": 0.0,
-                     "CRVAL2": 0.0,
-                     "CRPIX1": 2047.5,
-                     "CRPIX2": 2047.5,
-                     "CRPIX3": 0.0,
-                     "CDELT1": 0.0225,
-                     "CDELT2": 0.0225,
-                     "CDELT3": 1.0,
-                     "CUNIT1": "deg",
-                     "CUNIT2": "deg",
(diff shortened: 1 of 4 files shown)


punchbowl/data/tests/test_wcs.py
```diff
@@ -80,115 +80,246 @@ def test_extract_crota_helio():
...
+        eq_ra = eq_coords.ra.to_value(u.deg)
+        hp_to_eq_ra = hp_to_eq_coords.ra.to_value(u.deg)
+        straddles_wrap = np.nonzero(np.abs(eq_ra - hp_to_eq_ra) > 350)
+        for r, c in zip(*straddles_wrap):
+            if eq_ra[r, c] > 350:
+                eq_ra[r, c] -= 360
+            else:
+                hp_to_eq_ra[r, c] -= 360
+
+        np.testing.assert_allclose(eq_ra, hp_to_eq_ra, atol=1e-12)
```
svank, line 137: These are good lines! The coordinates computed from the equatorial WCS and those compute from the HP WCS and converted to equatorial agree to within 1e-12 degrees!

punchbowl/data/tests/test_wcs.py
```diff
@@ -80,115 +80,246 @@ def test_extract_crota_helio():
...
-def test_wcs_many_point_2d_check():
-    m = NormalizedMetadata.load_template("CTM", "2")
-    date_obs = Time("2024-01-01T00:00:00", format='isot', scale='utc')
-    m['DATE-OBS'] = str(date_obs)
+def test_helio_to_celestial_wcs_many_points():
+    for date_obs, crval1, crval2, crot in product(
+            ["2021-03-20T00:00:00", "2021-01-20T00:00:00"],
+            [-20, 10, 30],
+            [-10, 0, 10],
+            [-30, 0, 30] * u.deg):
```
svank, line 88: Looping over permutations like this could be done with `pytest.mark.parametrize`, but then 54 separate cases would show up in the test output!

svank:
This appears to fix our mosaic seams as well
image
image

jmbhughes (approved):
Good work!

PR #416, opened 2025-04-07
Adds ccd parameters as function input

lowderchris:
## PR summary

adds CCD parameters as inputs to sqrt decoding

## Todos

- [ ] determine if this is the way, or if all parameters should be read from meta


lowderchris:
Do we need this any longer? Or should we always take parameters from the FITS header?

jmbhughes:
Tasks should take it from the FITS header

PR #425, opened 2025-04-14
Square root decoding notebook

lowderchris:
## PR summary

- adds notebook demonstrating square root decoding
- adds notebook illustrating virtual environment setup


jmbhughes (approved):
As we discussed, in the environment setup please link to https://github.com/punch-mission/punch-mission?tab=readme-ov-file#setting-up-the-punch-environment and then this code is done.

PR #433, opened 2025-04-21
adds metadata to output jpeg2000 files

lowderchris:
## PR summary

adds metadata to output jpeg2000 files


jmbhughes (approved):
Once you add lxml to the dependency list, I'm happy to accept this.

PR #441, opened 2025-04-25
Two small fixes

svank:
## PR summary

Fixes a pydantic warning

Doesn't try to set a spacecraft location for F corona models. It doesn't make sense, and for the clears, CFMs are [marked as calibration files](https://github.com/punch-mission/punchbowl/blob/1283c68fdf788e28496b826c2e6a926f27c67d99/punchbowl/data/data/Level3.yaml#L88) which don't include the right header keys to store this data. Does this need a changelog?


jmbhughes (approved):
This needs a changelog update. But good work!

PR #454, opened 2025-05-12
Correct sign on p angle conversion

jmbhughes:
Ritesh noticed the F corona was angle in a wonky way. Plus the coordinate systems between celestial and helio weren't the same constellations. This should fix it (I think). 

@svank let's look at some files together using your plotting tool to make sure it's correct now.


svank (approved):
Needs a changelog?

PR #455, opened 2025-05-13
F-corona detrending, fixes for L2 metadata, and use header's gain and exposure time

svank:
## PR summary

- Detrending in F-corona estimation before outlier removal
- Read header values for gain and exposure time
- Enforce UTC for L2 date averaging, to avoid DST glitches
- Set FILEVRSN for L2s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 [Also available in GitHub](https://github.com/punch-mission/punchbowl/releases)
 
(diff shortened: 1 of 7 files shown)


jmbhughes (approved):
We should add threadpoolctl to the list of required packages (even though it's probably already a dependency of our dependencies). Other than that, this is good.

svank:
Looks like it's already in `pyproject.toml`

PR #456, opened 2025-05-14
Suppress CROTA warnings

svank:
## PR summary

This is a high-priority fix: if we've decided to ship FITS files that raise Astropy warnings, we should suppress them in our own code.

## Breaking changes

The `CROTA` warnings were breaking me


lowderchris (approved):
Warning : These changes look alright.

PR #468, opened 2025-05-19
promotes data earlier in level 1 flow

lowderchris:
## PR summary

promotes data earlier in level 1 flow, to allow for calibration metafields


jmbhughes:
Contained in a different PR.

PR #471, opened 2025-05-21
adds module and test for flagging in uncertainty of saturated pixels

lowderchris:
## PR summary

For data pixels that are saturated, the uncertainty layer is now flagged as infinite uncertainty.


jmbhughes (commented):
I wonder if the saturation value should come from the header and be set earlier. We might run into inconsistency if we calculate it on the fly every time.

lowderchris:
> I wonder if the saturation value should come from the header and be set earlier. We might run into inconsistency if we calculate it on the fly every time.

Absolutely - should be done now. The corresponding DSATVAL should already be set by punchpipe (https://github.com/punch-mission/punchpipe/blob/46fb2073917f1891fc0127295e1dd36e4428bf7f/punchpipe/flows/level0.py#L752)

jmbhughes (approved):

PR #495, opened 2025-06-19
add new documentation

jmbhughes:
Adds documentation about data versions and anomalies.


jmbhughes:
The CI is failing for an unrelated reason. I opened #496 to fix that. But it doesn't impact this PR.

PR #508, opened 2025-06-27
Don't copy certain meta keys from L0 to L1

### LINKED ISSUE #504: Level 1 files have ISSQRT flag set
Should not be included or it's misleading

svank:
## PR summary

We were copying a lot of data from the L0 headers into the L1 headers, but some of that should have been left behind. We were generating L1s with `TITLE   = 'PUNCH Level-0 NFI-0 Clear Image'`, `BUNIT   = 'sqrt(DN)'`, etc. (This was overriding the correct default in the L1 omniheader.)

These are just the keys that stood out to me, it's possible there are more we shouldn't copy.

Closes #504

### DIFF
--- a/punchbowl/level1/flow.py
+++ b/punchbowl/level1/flow.py
@@ -169,6 +169,8 @@ def level1_core_flow(  # noqa: C901
         new_meta = NormalizedMetadata.load_template(product_code, "1")
         # copy over the existing values
         for key in data.meta.keys(): # noqa: SIM118
+            if key in ["BUNIT", "DESCRPTN", "FILENAME", "ISSQRT", "LEVEL", "TITLE"]:
+                continue
             if key in new_meta.keys(): # noqa: SIM118
                 new_meta[key] = data.meta[key].value
         new_meta.history = data.meta.history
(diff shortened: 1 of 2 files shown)


punchbowl/level1/flow.py
```diff
@@ -169,6 +169,8 @@ def alignment_mask(x: float, y: float) -> bool:
         new_meta = NormalizedMetadata.load_template(product_code, "1")
         # copy over the existing values
         for key in data.meta.keys(): # noqa: SIM118
+            if key in ["BUNIT", "DESCRPTN", "FILENAME", "ISSQRT", "LEVEL", "PIPEVRSN", "TITLE"]:
```
lowderchris, line 172: Could we make this an updatable list at the top?

lowderchris (approved):
Looks good. Ruff is failing because of that __hash__ issue that we were seeing earlier. Made one suggestion about making a list of skippable keywords if that works better.

punchbowl/level1/flow.py
jmbhughes: I agree.

jmbhughes (approved):

punchbowl/level1/flow.py
```diff
@@ -187,8 +191,7 @@ def alignment_mask(x: float, y: float) -> bool:
         if isinstance(quartic_coefficient_path, Callable):
             _, quartic_coefficient_path = quartic_coefficient_path()
         new_meta["CALCF"] = os.path.basename(quartic_coefficient_path) if quartic_coefficient_path else ""
-        new_meta["LEVEL"] = "1"
-        new_meta["FILEVRSN"] = data.meta["FILEVRSN"].value
+        new_meta["FILEVRSN"] = "0a"
```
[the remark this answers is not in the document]
jmbhughes: https://github.com/punch-mission/punchbowl/issues/511

PR #529, opened 2025-07-21
NFI LQ batching and improved outlier detection

svank:
This and the punchbowl side ended up being larger than I anticipated, but it works!

* Adds `Limit` and `LimitSet` classes, which represent outlier detection levels which are functions of a variable (e.g. a `DATAMDN` vs. `CROTA` curve). This gives more robust rejection than just setting limits on a single variable (e.g. a fixed `DATAMDN` threshold). Includes serialization and some convenience code for plotting the limits.
* Allows batches of NFIs to be PCAd and subtracted. Files are omitted from fitting and then subtracted on a staggered basis, and the stagger positions run simultaneously.
* Extra images that are just for PCA fitting, not subtraction, are not excluded by the stagger, to give better-quality PCA fits when we're not doing a full batch of 1000 images-to-be-subtracted (i.e. when we're not reprocessing)
* The images-to-be-subtracted are loaded in parallel, which saves a few minutes for a batch of 1000.
* The subtracted images are reprojected in parallel, using Prefect task parallelism and Task concurrency limits
* I threw in one extra trick I used for The Video. We're using a quartering approach to prevent the Moon and planets from contaminating the PCA components. That sometimes leaves seams at the quarter boundaries, so now there's 100 px of overlap between quarters and a linear fade from one quarter to another, to smooth the seams.

TODO, whether here or as a follow-on:

- [ ] Be more intentional about outlier images-to-be-subtracted. Right now they're both excluded from fitting and not subtracted, but we maybe still want to attempt subtraction.
- [ ] Try excluding masked regions from the PCA. This could be a 10-20%-ish speedup


svank:
pre-commit.ci autofix

svank:
Added exclusion of the masked region from the PCA. Trimmed the end-to-end flow runtime from 58 to 51.5 minutes for a 1000-file batch. (BTW if that runtime is too much latency, we can try adjusting things to cut it down. I have ideas)

punchbowl/levelq/flow.py
```diff
@@ -1,40 +1,57 @@
...
 
 ORDER_QP = ["CR1", "CR2", "CR3", "CNN"]
 
 @flow(validate_parameters=False)
 def levelq_CNN_core_flow(data_list: list[str] | list[NDCube], #noqa: N802
                          output_filename: list[str] | None = None,
-                         files_to_fit: list[str | NDCube | Callable] | None = None) -> list[NDCube]:
+                         files_to_fit: list[str | NDCube | DataLoader] | None = None,
+                         outlier_limits: str | None = None,
+                         data_root: str | None = None) -> list[NDCube]:
```
jmbhughes, line 25: We really should have a more detailed doc string that describes these things. I know you inherited it like this, but could you fix it?

punchbowl/levelq/limits.py
```diff
@@ -0,0 +1,172 @@
...
+            x = point[self.xkey].value
+            y = point[self.ykey].value
+        elif isinstance(point, Iterable) and isinstance(point[0], (Header, NormalizedMetadata)):
+            return np.array([self.is_good(p) for p in point])
+        else:
+            x, y = point
+
+        limit_value = np.interp(x, self.xs, self.ys)
+
+        if self.comp == "<":
```
jmbhughes, line 57: A match statement might be less verbose here

punchbowl/levelq/limits.py
```diff
@@ -0,0 +1,172 @@
+from collections.abc import Iterable
```
jmbhughes, line 1: @d-lamb you might find @svank 's implementation of limits helpful for outlier rejection at raw --> L0

punchbowl/levelq/pca.py
```diff
@@ -11,114 +18,203 @@
...
-    for i, quarter_slice in enumerate(quarter_slices):
-        logger.info(f"Starting to filter quarter {i+1}")
-        no_bodies_in_quarter = np.all(bodies_in_quarter[:, :, i] == False, axis=1) # noqa: E712
-        images_for_quarter = all_files_to_fit[no_bodies_in_quarter]
-        filtered = run_pca_filtering(input_cube, images_for_quarter, n_components, med_filt)
+    try:
+        # We stash the loaded images in a global variable so that, when we fork for parallel processing,
+        # those images don't have to be individually copied to each worker process. The try/finally block ensures we
+        # don't leak this memory.
+        global _all_files_to_fit # noqa: PLW0603
```
jmbhughes, line 45: We do what we have to do.

punchbowl/levelq/pca.py
```diff
@@ -11,114 +18,203 @@
...
-    logger.info(f"Kept {index_to_insert}, filling {all_files_to_fit.nbytes / 1024 ** 3:.2f} GB")
+    logger.info(f"Total of {len(all_files_to_fit)} images to fit, filling {all_files_to_fit.nbytes / 1024 ** 3:.2f} GB")
+    logger.info(f"(Drawn from {len(input_cubes)} images to subtract and {len(files_to_fit)} extra images for fitting)")
+    logger.info(f"({n_outliers} outliers were rejected from fitting)")
 
-    bodies_in_quarter = np.array(bodies_in_quarter)
+    logger.info("Locating planets")
+    # We have a lot of data in memory right now, so forking is expensive as all that memory has to be marked as
+    # copy-on-write. Using a forkserver avoids that work.
+    ctx = mp.get_context("forkserver")
```
jmbhughes, line 138: This definitively means we don't support Windows, right? It's only supported on POSIX systems.

punchbowl/levelq/pca.py
```diff
@@ -11,114 +18,203 @@
...
 
-        reconstructed = pca.inverse_transform(transformed).reshape(input_cube.data.shape)
-        return input_cube.data - reconstructed
+    reconstructed = np.zeros_like(images_to_subtract)
+    reconstructed[:, masked_region] = pca.inverse_transform(transformed)
+    return images_to_subtract - reconstructed
 
 
-def find_bodies_in_image(frame: str | NDCube | WCS) -> list:
+def find_bodies_in_image(frame: str | NDCube | tuple[NormalizedMetadata, WCS], extra_padding: int = 0) -> list:
```
jmbhughes, line 217: This seems more general purpose. We might move this `find_bodies_in_image` function to somewhere like `punchbowl.data.wcs`

punchbowl/levelq/pca.py
```diff
@@ -129,15 +225,20 @@ def find_bodies_in_image(frame: str | NDCube | WCS) -> list:
...
+    else:
         msg = "Type of 'frame' not recognized"
         raise TypeError(msg)
 
     results = []
     for body in ["Mercury", "Venus", "Moon", "Mars", "Jupiter", "Saturn"]:
         body_loc = get_body(body, time=Time(wcs.wcs.dateobs), location=EarthLocation.from_geodetic(*location))
         x, y = wcs.world_to_pixel(body_loc)
-        w = 40 if body == "Moon" else 10
+        w = 100 if body == "Moon" else 10
```
jmbhughes, line 240: The logic here being that the moon is big? Could you document this number with a comment?

punchbowl/levelq/pca.py
```diff
@@ -149,3 +250,49 @@ def find_bodies_in_image(frame: str | NDCube | WCS) -> list:
...
     return results
+
+
+@contextlib.contextmanager
+def _log_forwarder() -> None:
+    # This logging situation is kind of a mess. We really want to parallelize with multiprocessing so we can fork and
+    # not copy all the images to each worker. But Prefect logging from those forked processes doesn't work (nor does
+    # logging print statements), so in the workers we need to set up a logger than puts log entries in a
+    # shared-memory Queue, and then on the main process side we need to spawn a thread that monitors the queue and
+    # forwards log entries to Prefect.
```
jmbhughes, line 261: Python is the best.

punchbowl/levelq/pca.py
svank: Maybe, but this specifically marks which quarters of the image the planets are in, which feels less general.

svank:
I know I just pushed, but let's let a flow run successfully on 190 before this PR goes anywhere. Also still todo, tests and docs

svank:
OK, this now at least attempts a subtraction for outlier images, while still excluding them from the PCA fitting. This is also free of mystery hangs, swallowed exceptions, etc. There are tests, though coverage doesn't track what the worker processes run.

Only remaining todo is your documentation request which I'll do tomorrow, but this code is ready to go.

PR #536, opened 2025-07-23
separates stray light calibration logging

lowderchris:
Separates stray light calibration logging to note both before and after files used


jmbhughes:
@lowderchris @svank this PR might need to change depending on what we do with stray light modeling. Maybe just the comment because it won't necessarily be a "before" and "after" model.

PR #538, opened 2025-07-24
Misc improvements

svank:
* Stray light: set date_obs to the reference time, and cut memory use significantly
* Cache layer: use the `DataLoader` API everywhere---it's cleaner
* Log spam: only set the debugging `on_completion` hook if debug is enabled. This flow set-up code runs afresh in the new process every time a flow starts, so the debug state can still be changed live. But now we don't get log messages about the `on_completion` hook in non-debug mode.


jmbhughes (approved):
I'll switch the changelog over and merge this.

PR #556, opened 2025-08-05
updates data querying notebook

lowderchris:
streamlines data querying notebook to:

query a shorter span of time
only retrieve the first result
removes SunPy < 7.0 workarounds


lowderchris:
Read the docs is having issues stemming from the Fido VSO search intermittently returning no files.

PR #557, opened 2025-08-06
Change NFI pointing alignment to work differently than WFI

jmbhughes:
NFI pointing alignment was failing because of astrometry.net's initial solver. This was in part to incorrect search scales and also because of weird "non-stars" being detected as stars. We now try to use astrometry.net but if it fails we warn and proceed with the instrument WCS.


jmbhughes:
pre-commit.ci autofix

lowderchris (approved):

PR #567, opened 2025-08-12
NFI flat-fielding

lowderchris:
## PR summary

Integrates NFI flat-fielding code into punchbowl

## Todos

- [ ] further testing with calibration data?
- [ ] integration with punchpipe

### DIFF
--- a/punchbowl/level1/vignette.py
+++ b/punchbowl/level1/vignette.py
@@ -193,6 +193,8 @@ def generate_vignetting_calibration_nfi(input_files: list[str],
                                         dark_path: str,
                                         path_mask: str,
                                         polarizer: str,
+                                        dateobs: str,
+                                        version: str,
                                         output_path: str | None = None) -> np.ndarray | None:
     """
     Create calibration data for vignetting for the NFI spacecraft.
@@ -207,6 +209,10 @@ def generate_vignetting_calibration_nfi(input_files: list[str],
         Path to the speckle mask FITS file
     polarizer : str
         Polarizer name
+    dateobs : str
+        Timestamp for calibration file
+    version : str
+        File version
     output_path : str | None
         Path to calibration file output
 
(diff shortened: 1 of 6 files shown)


punchbowl/level1/vignette.py
```diff
@@ -178,6 +185,118 @@ def generate_vignetting_calibration(path_vignetting: str,
...
+
+    # Load a WCS to use later on
+    with fits.open(input_files[0]) as hdul:
+        cube_wcs = WCS(hdul[1].header)
+
+    # Load and square root decode input data
+    cubes = [
+        decode_sqrt_data.fn(cube)
+        for cube in (load_ndcube_from_fits(file) for file in input_files)
+        if 490 <= cube.meta["DATAMDN"].value <= 655 and cube.meta["DATAP99"].value != 4095
```
jmbhughes, line 238: I assume these numbers are set by experience. Will they ever need to change? Do we have any rationale for them?

punchbowl/level1/vignette.py
```diff
@@ -178,6 +185,118 @@ def generate_vignetting_calibration(path_vignetting: str,
...
+    mask_nfi = load_spacecraft_mask(path_mask)
+
+    # Deal with infs and remask
+    nfiflat[np.isinf(nfiflat)] = 1.
+    nfiflat[mask_nfi == 0] = 1
+
+    # Generate an output metadata and NDCube
+    m = NormalizedMetadata.load_template(f"G{polarizer}4", "1")
+    # TODO - set these dynamically, or from passed in variables
+    m["DATE-OBS"] = "2025-07-31T00:00:00.000"
```
jmbhughes, line 275: i don't think we always want this to be the date-obs.

punchbowl/level1/vignette.py
```diff
@@ -178,6 +185,118 @@ def generate_vignetting_calibration(path_vignetting: str,
...
+    m["FILEVRSN"] = "0e"
```
jmbhughes, line 276: This version will need to change depending on the config.

jmbhughes (changes requested):

punchbowl/level1/vignette.py
lowderchris: This might be a question for @s0larish

s0larish: Yes as said by Marcus, it was based on looking at the DATAMDN value over several days of data and identifying the optimum range

jmbhughes (approved):

PR #578, opened 2025-08-15
separates mask paths

lowderchris:
separates mask paths for speckle masks versus instrument masks

TODO

- [x] Confirm OFFSET setting

### DIFF
--- a/changelog/586.bugfix.rst
+++ b/changelog/586.bugfix.rst
@@ -1,0 +1,1 @@
+DATE is set correctly in L1 headers.
--- a/changelog/592.feature.rst
+++ b/changelog/592.feature.rst
@@ -1,0 +1,1 @@
+Rolls off the uncertainty by distance to edge of mask so that we seamlessly make mosaics.
--- a/changelog/593.doc.rst
+++ b/changelog/593.doc.rst
@@ -1,0 +1,1 @@
+Update the data version documentation.
--- a/changelog/595.bugfix.rst
+++ b/changelog/595.bugfix.rst
@@ -1,0 +1,1 @@
+Updates quickpunch f-corona generation for multithreading and data cube array ordering.
--- a/changelog/597.feature.rst
+++ b/changelog/597.feature.rst
@@ -1,0 +1,1 @@
+Add spacecraft data example notebooks in punchbowl/examples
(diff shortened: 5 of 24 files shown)


s0larish (approved):

lowderchris:
The last check would be to check that this block:
```
    cubes = [
        decode_sqrt_data.fn(cube)
        for cube in (load_ndcube_from_fits(file) for file in input_files)
        if cube.meta["OUTLIER"].value == 0
    ]
```
is equivalent to replace,
```
    cubes = [
        decode_sqrt_data.fn(cube)
        for cube in (load_ndcube_from_fits(file) for file in input_files)
        if 490 <= cube.meta["DATAMDN"].value <= 655 and cube.meta["DATAP99"].value != 4095
           and not cube.meta.__setitem__("OFFSET", 400)
    ]
```

svank:
I think it won't be an exact replacement, but it's the same intent and the `OUTLIER` flag *should* be pretty reliable.

(I'm confused about the `__setitem__` in the old code)

punchbowl/level1/vignette.py
```diff
@@ -256,6 +264,10 @@ def generate_vignetting_calibration_nfi(input_files: list[str],
 
     # Stack image data
     images = np.array([cube.data for cube in cubes])
+
+    if len(images.shape) != 3:
+        return None
```
jmbhughes, line 269: Why do there have to be 3?

jmbhughes: Please just add a comment.

punchbowl/level1/vignette.py
```diff
@@ -283,7 +295,7 @@ def generate_vignetting_calibration_nfi(input_files: list[str],
     cube = NDCube(data=nfiflat.astype("float32"), wcs=cube_wcs, meta=m)
 
     if output_path is not None:
-        filename = f"{output_path}/{get_base_file_name(cube)}.fits"
+        filename = f"{output_path}{get_base_file_name(cube)}.fits"
```
jmbhughes, line 298: You don't need the "/"?

jmbhughes (changes requested):

jmbhughes (approved):

PR #586, opened 2025-08-28
Set DATE in L1 files

svank:
## PR summary

We were just copying the value from the L0 file.


svank:
At least the first test failure is incompatible units: `<ufunc 'add'>, unit1 = Unit("deg"), unit2 = Unit("deg2")`

PR #606, opened 2025-10-05
Add PIM and CIM data codes

jmbhughes:
Adds the PIM and CIM products to the data codes table


jmbhughes:
pre-commit.ci autofix

PR #647, opened 2025-10-21
Adds noise reconstitution

jmbhughes:
We always need to add back in the noise we lost with sqrt encoding. 

- [x] need to set the seed correctly

### DIFF
--- a/punchbowl/level1/sqrt.py
+++ b/punchbowl/level1/sqrt.py
@@ -114,12 +114,13 @@ def decode_sqrt(
         out_data = decode_sqrt_by_table(data, table_bottom)
 
     if reconstitute_noise:
+        scale = np.sqrt(out_data) / 2.2
         if dateobs is None:
-            noise = np.random.normal(scale=np.sqrt(out_data) / 2.2)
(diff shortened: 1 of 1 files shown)


punchbowl/level1/sqrt.py
```diff
@@ -106,6 +113,14 @@ def decode_sqrt(
...
         out_data = decode_sqrt_by_table(data, table_bottom)
 
+    if reconstitute_noise:
+        if dateobs is None:
+            noise = np.random.normal(scale=np.sqrt(out_data) / 2.2)
+        else:
+            seed = int(dateobs.timestamp())
+            rng = np.random.default_rng(seed)
+            noise = rng.normal(scale=np.sqrt(out_data) / 2.2)
+        return out_data + noise
```
svank:
```suggestion
        scale = np.sqrt(out_data) / 2.2
        if dateobs is None:
            noise = np.random.normal(scale=scale)
        else:
            seed = int(dateobs.timestamp())
            rng = np.random.default_rng(seed)
            noise = rng.normal(scale=scale)
        return out_data + noise
```

svank (approved):

PR #664, opened 2025-11-03
Mosaic fixes

svank:
* Fixes use of `conserve_flux`
* Fix detection of image edges for cropping and rolloffs, and make the rolloff configurable


jmbhughes (approved):
Great stuff!

PR #671, opened 2025-11-17
Sets zeros in datacubes to NaNs  

s0larish:
Sets zeros in datacubes to NaNs before estimating the percentile and making the polarized f corona background


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #680, opened 2025-12-09
Use a percentile band for F-corona generation

svank:
Rather than using every image whose tB is under a threshold, let's use the images who are between two thresholds.


svank:
pre-commit.ci autofix

jmbhughes (approved):

PR #681, opened 2025-12-09
Two small changes

svank:
Add an option to set the dtype of a loaded NDCube. (When I'm loading a lot of them at once in my notebooks, sometimes I'd rather use `float32` and save lots of memory.

Give `NormalizedMetadata` a `__str__` so we can `print(meta)` and see the full header.


svank:
pre-commit.ci autofix

PR #702, opened 2026-01-06
Add more diagnostic data to headers

svank:
We had talked about adding more server state data to the headers---how does this look? I think I'd set them all in punchpipe so it's a one-and-done thing rather than an addition to every single flow.


svank:
I should test this out more, but what about this? I had to add the new section to the level definition files, so I also tried to exclude versions of packages that don't factor into making a given file.

jmbhughes (approved):
Thanks I like that better.

PR #706, opened 2026-01-16
Don't crash on import if Prefect can't be reached

svank:
The `hasten-db` tests are now failing in CI but not locally. It's failing with a "can't reach ephemeral Prefect server` error that traces back to this line. If I run locally without Prefect running I get a different error on this line, and this change stops it, so let's see if this helps CI?

Not sure why this wasn't a problem before...

### DIFF
--- a/changelog/706.bugfix.rst
+++ b/changelog/706.bugfix.rst
@@ -1,0 +1,1 @@
+Fix for possible crash on import if a Prefect server isn't running.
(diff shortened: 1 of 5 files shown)


svank:
Looks like CI already had some failures---hoping this pushes fixes them!

jmbhughes (approved):

PR #708, opened 2026-01-21
Attempt to clean git history

jmbhughes:
We have some annoying FITS files in the git history and some other large files. Our total repo size is > 800 MB so this attempts to clean that up.


jmbhughes:
This makes sense. I hadn't looked into it closely. I figured a PR wouldn't be the solution, but I wanted to raise the issue here rather than destroying the repo without discussion. 😄

Issue #674, opened 2025-11-25
OBSLAYR in L3 PIMs


svank:
In `Level3.yaml`, PIMs are of the `Polarized` kind, which omits OBSLAYR3. But at least currently, our PIMs are still in MZP, so they get written with `OBSLAYR1= 'Polar_M '` and `OBSLAYR2= 'Polar_Z '`, but no `OBSLAYR3` key

PR #715, opened 2026-01-23
Fix obslayer

s0larish:
Fixed OBSLAYR3 issue in L3 files #674


jmbhughes (commented):
I'd like to see the generated headers for level 3 products so we can verify the correctness of this.

jmbhughes (approved):
Received headers via Mattermost

PR #722, opened 2026-01-28
Streamline example notebooks

lowderchris:
Streamlines example notebooks built as part of the documentation to avoid timing out the CI.
- Swaps RGB map notebook to use staged sample data
- Removes data animation notebook from the dynamic build for now


jmbhughes (changes requested):
Your new notebook doesn't actually show the video. 

Image

Also, the RGB image looks pretty bad. 

Image

jmbhughes (approved):

PR #760, opened 2026-01-31
Added metadata to indicate if the moon is in the PUNCH FOV

s0larish:
## PR summary

Adds code to forecast/estimate when Moon will appear in PUNCH field of view.


jmbhughes:
pre-commit.ci autofix

changelog/760.feature.rst
jmbhughes: I renamed this so it conforms to our style.

punchbowl/data/tests/test_meta.py
```diff
@@ -300,3 +306,12 @@ def test_fits_header_has_both_distortion_models():
     assert "DP2" in h
     assert "DP1A" in h
     assert "DP2A" in h
+
+
+def test_check_moon_in_fov():
```
jmbhughes: This tests the function but not the metadata.

punchbowl/data/meta.py
```diff
@@ -892,3 +893,67 @@ def set_spacecraft_location_to_earth(input_data: NDCube) -> NDCube:
...
+    time_start = Time(time_start_str)
+    time_end = Time(time_end_str)
+    dt = TimeDelta(resolution_minutes * 60, format="sec")
+    times = time_start + dt * np.arange(int((time_end - time_start) / dt) + 1)
+    angles = []
+    in_fov_times = []
+
+    fov_half = fov_deg / 2
+
+    with solar_system_ephemeris.set("jpl"):
```
jmbhughes: This seems to be triggering an error in the CI.

punchbowl/data/meta.py
```diff
@@ -892,3 +893,67 @@ def set_spacecraft_location_to_earth(input_data: NDCube) -> NDCube:
...
+            # Compute angle
+            dot = np.dot(v_moon_unit, v_sun_unit)
+            angle = np.degrees(np.arccos(np.clip(dot, -1.0, 1.0)))
+
+            angles.append(angle)
+
+            if angle <= fov_half:
+                in_fov_times.append((t.iso, angle))
+
+    if plot:
```
jmbhughes: I think I'd like to remove the plotting routine. Functions are better if they do just one thing. We could have a separate plotter if you'd like, but not in this function.

jmbhughes (changes requested):
My main request is to finish this so it puts the moon distance in the header of our Level 0 and Level 1 FITS files. 

I did fix a couple things and make more comments so you have guidance along the way.

jmbhughes:
pre-commit.ci autofix

punchbowl/level2/flow.py
```diff
@@ -164,6 +164,13 @@ def level2_core_flow(data_list: list[str] | list[NDCube], # noqa: C901
     output_data.meta["DATE-END"] = output_dateend
     output_data = set_spacecraft_location_to_earth(output_data)
 
+    _, angle_sun, _, _, _, xpix, ypix = check_moon_in_fov(
+        output_data.meta.meta["DATE-OBS"].value, wcs=output_data.wcs,
```
jmbhughes: You have an extra `.meta` here

jmbhughes (changes requested):
I think this is why the CI is failing.

punchbowl/level2/flow.py
s0larish: Thanks :)

jmbhughes (changes requested):
Please fix the merge conflict and add a test showing the lunar meta is correct after a flow test, i.e. is it actually in the file?

jmbhughes (commented):
Why is the omniheader showing everything changed in the diff? I don't want to do something on accident because that hides the true changes.

jmbhughes (approved):

PR #782, opened 2026-02-13
Exclude bad-packet images from certain steps

svank:
Bad packet images have a *lot* of pixels that look saturated. For these images, we shouldn't despike, destreak, or fill saturated pixels. (If nothing else, it just takes a long time to run with that many "saturated" pixels.)


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #795, opened 2026-02-17
Big batch of stray light fixes

svank:
This got the new stray light modelling to work!


svank:
pre-commit.ci autofix

jmbhughes (dismissed):

svank:
The last failing test is testing that the SL model generation runs. Fixing that up might be a bit of work, and the model generation is a lot more CPU intensive now, so it might be a longer-running test. Since we're anticipating re-writing it again soon, what do you think about just `xfail`ing it for now?

jmbhughes (approved):

PR #800, opened 2026-02-19
Don't crash in eclipse season

svank:
Every stray light model we generate is passed a fallback model, and if a given CROTA bin has too many outliers, the model copies that bin from the fallback model. It doesn't make pretty images, but it lets things run until we make something better for 0k.


jmbhughes (approved):

jmbhughes:
pre-commit.ci autofix

PR #804, opened 2026-02-23
Remove redundancy in imax effect correction

s0larish:
## PR summary
IMAX effect is being taken care off in solpolpy new function of mzpinstru_to_mzpsolar accounting for solar north


jmbhughes (changes requested):
This needs an update @s0larish based on your solpolpy changes.

PR #806, opened 2026-02-25
Data visualization bug fixes

lowderchris:
Fixes a bug with array indexing for multi-dimensional data, and improves memory handling for large animations.

### DIFF
--- a/punchbowl/data/visualize.py
+++ b/punchbowl/data/visualize.py
@@ -1,5 +1,6 @@
+import gc
+import warnings
 import subprocess
-from copy import deepcopy
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from multiprocessing import Pool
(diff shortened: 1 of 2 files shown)


punchbowl/data/visualize.py
```diff
@@ -17,6 +18,7 @@
 
 from punchbowl.data import punch_io
 
+warnings.filterwarnings("ignore", message="invalid value encountered in do_format", category=RuntimeWarning)
```
jmbhughes: One question: what causes this warning that we're ignoring?

jmbhughes (approved):

punchbowl/data/visualize.py
lowderchris: This was cropping up when hacking those movies together, and a search indicated it could be relaxed to axes coordinate handling. However, when I ran this again locally on a clean branch, I couldn't reproduce the warning. So I'll remove it for now.

jmbhughes (commented):
Please think carefully about using gc every iteration of the for loop.

PR #808, opened 2026-02-26
Add missing comma, fix logging in L3

jmbhughes:
A missing comma in the cleaner resulted in unintended resetting of files.


svank (approved):
There are 1,077,294 characters in our `*.py` files, so this comma was literally one millionth of our codebase. How much of an impact could this have?

PR #813, opened 2026-03-04
Update QuickPUNCH products to new definitions

jmbhughes:
We've redefined the quickpunch products. This carries out the changes. It includes renaming LQ_CNN to LQ_QNN and adding a LQ_QAM product. The resolutions are now the same as regular PUNCH products instead of 1k x 1k. 

This also begins writing SHA hashes for JP2 files.


lowderchris (approved):
Made two changes (revert if they don't seem alright), but otherwise it all seems good.

PR #821, opened 2026-03-11
L2 masking

svank:
## PR summary

Allows per-WFI masks to be applied in L2 before merging. I'm having them be `PUNCH_L2_MS*.bin`, vs. the `PUNCH_L1_MS*.bin` files that we apply in L1. The mask application sets the masked regions to `nan`, so `reproject` registers them as values not-to-be-sampled.

This was tested in a 190 pipeline run, but only for WFI clears.


jmbhughes (changes requested):
Can you add an entry in the product codes list in the docs? Also some tests are failing.

PR #869, opened 2026-04-09
Allows setting of dtype in polarization conversion

lowderchris:
Allows setting of dtype in polarization conversion, to assist in polarized starfield model generation.

### DIFF
--- a/changelog/869.bugfix.rst
+++ b/changelog/869.bugfix.rst
@@ -1,0 +1,1 @@
+Allows setting of dtype in polarization conversion, to assist in polarized starfield model generation.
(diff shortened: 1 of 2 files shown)


lowderchris:
@svank This will hopefully fix the witty-tanuki polarized starfield model that just failed.

Also, I'm not sure what the deal is with the CI - lots of Prefect connection failures.

lowderchris:
And now the CI is passing. Go figure.

jmbhughes (changes requested):
You sent the default to dtype=None. What does that do? I assume it would fail if ran like that. Can you default to 32 bit floats?

jmbhughes (approved):

PR #909, opened 2026-04-22
Pass max_workers through to alignment

svank:
Maybe we won't need this later, but for now we can pass through max_workers to the alignment step. (I'm running l1_early in a notebook for the Artemis search, and I want to use more workers!)


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #912, opened 2026-04-24
Uncompressed movie scaling

### LINKED ISSUE #911: Movies of non-square-root-compressed data are saturated
**Describe the bug**
The scaling of the color table is incorrect for the non-square-root-compressed WFI data taken starting 2026-04-20 after the LED calibration sequence.

**To Reproduce**
1. View the movie L0_CR1_2026-04-20T00/00/29.303000+00/00-2026-04-20T03/28/29+00/00.mp4

**Expected behavior**
The images in the movie should look similar before and after the break for the LED sequence.

**Screenshots**

Image

Image

d-lamb:
## PR summary

Scale uncompressed data properly when making the daily quicklook movies

## Test plan

*How does this PR verify the code works correctly?*
See saturated image in #911.
This is what happens when the Level==0 is changed to Level=="0":
Squared VminVmax
This is what happens when the scaling is applied after the squaring:
Squared-and-Scaled VminVmax

## Breaking changes

None known. I have not been able to test this on uncompressed NFI data.

## Related Issues

Fixes #911.

### DIFF
--- a/changelog/912.bugfix.rst
+++ b/changelog/912.bugfix.rst
@@ -1,0 +1,1 @@
+When non-square-root-compressed data is acquired, appropriately scale the color bar range in the daily movies.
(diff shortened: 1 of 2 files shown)


lowderchris (approved):
Good catch on the string versus int level.

PR #913, opened 2026-04-24
Support an outlier-binary color choice in the dashboard

svank:
In the files dashboard, now that the `outlier` flag can be a bitmask, coloring the points by outlier status gives a gradient from 0 to 14, in which 1 looks a lot like 0. This adds an option that colors the points by `outlier != 0`


jmbhughes (approved):

jmbhughes:
pre-commit.ci autofix

PR #924, opened 2026-05-03
New faster alignment code

jmbhughes:
Not quite finished with PR


jmbhughes:
I'm going to kill this PR since the algorithm was not as accurate as desired. @svank figured out a way to speed up the existing algorithm. Yay!

Issue #687, opened 2025-12-14
Highlight the data codes that users will most likely want


svank:
We should make a friendly guide to the most useful data products. Maybe even a "If you want to do this, try using this"

PR #925, opened 2026-05-04
Data description update 

s0larish:
## PR summary

Update the documentation to highlight which dataproducts to be used primarily for scientific analysis.
Completes #687


docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
 Accessing PUNCH Data
 ====================
 
+Which Product to look for?
+--------------------------
+Recommended Level 3 Products
+-----------------------------
+
+For most science use cases, the recommended starting point is the Level 3 **CAM**
+(Clear a.k.a total brightness low-noise science mosaic) and **PAM** (Polarized low-noise science mosaic) products.
```
lowderchris: I might flip this throughout the documentation - so refer to them first as clear / polarized low-noise mosaics, with CAM / PAM in parentheses? Just to avoid getting lost in an alphabet soup of product codes for users.

docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
...
+giving you the cleanest possible signal for heliospheric science.
+CAM provides total brightness (B) with uncertainty, while PAM additionally resolves
+polarized brightness (pB) and its radial component (pB'), making PAM the product of
+choice for studies of the CMEs, shocks, solar wind etc. These products
+represent one full spacecraft rotation cycle with a significant
+improvement in signal-to-noise relative to other products.
+
+Researchers who need intermediate data products — for example to inspect F-corona
+subtraction before the final background removal, or to carry out their own starfield
+subtraction — should consider the **CIM** (Clear/total brightness science trefoil with F-corona subtraction)
```
jmbhughes: We no longer ship CIM and PIM to the SDAC, so it's not worth mentioning.

docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
...
+and **PIM** (Polarized science trefoil with F-corona subtraction) products. CIM and PIM
+retain the trefoil geometry of the full PUNCH field of view but stop short of the full
+background subtraction pipeline applied to CAM and PAM. They are particularly useful for
+validating custom reduction steps or for science cases that are sensitive to how the
+background model is constructed.
+
+If you are unsure which product to start with, download a CAM or PAM file first.
+Use CIM/PIM only if your analysis requires it.
```
jmbhughes: Same again for CIM/PIM.

docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
...
 ====================
 
+Which Product to look for?
+--------------------------
+Recommended Level 3 Products
+-----------------------------
+
+For most science use cases, the recommended starting point is the Level 3 **CAM**
+(Clear a.k.a total brightness low-noise science mosaic) and **PAM** (Polarized low-noise science mosaic) products.
+Both are fully background and starfield subtracted and mosaicked across the PUNCH field of view,
```
jmbhughes:
```suggestion
Both products are fully background and starfield subtracted and mosaicked across the PUNCH field of view,
```

docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
 Accessing PUNCH Data
 ====================
 
+Which Product to look for?
+--------------------------
+Recommended Level 3 Products
+-----------------------------
```
jmbhughes: Are there two headings here?

docs/data/access.rst
svank: But on the other hand, if we focus on "polarized" and "clear", that might distract from the fact that PAM also includes unpolarized data.

docs/data/access.rst
```diff
@@ -1,6 +1,33 @@
...
+
+For most science use cases, the recommended starting point is the Level 3 **CAM**
+(Clear a.k.a total brightness low-noise science mosaic) and **PAM** (Polarized low-noise science mosaic) products.
+Both are fully background and starfield subtracted and mosaicked across the PUNCH field of view,
+giving you the cleanest possible signal for heliospheric science.
+CAM provides total brightness (B) with uncertainty, while PAM additionally resolves
+polarized brightness (pB) and its radial component (pB'), making PAM the product of
+choice for studies of the CMEs, shocks, solar wind etc. These products
+represent one full spacecraft rotation cycle with a significant
+improvement in signal-to-noise relative to other products.
```
svank:
```suggestion
For most science use cases, the recommended starting point is the two Level 3 low-noise science mosaic products,
**PAM** and **CAM**.
Both products are fully background and starfield subtracted and mosaicked across the PUNCH field of view,
giving you the cleanest possible signal for heliospheric science.
Each of these two product types are produced every 32 minutes, allowing multiple images to be averaged together at every point in the mosiac field of view,
offering improved signal-to-noise relative to other products.
PAM files (with the "P" standing for "polarized") have three data layers: total brightness (tB), which is effectively an unpolarized image, followed by
polarized brightness (pB) and its radial component (pB'), making PAM the product of
choice for studies of CMEs, shocks, the solar wind, etc.
CAM files (with the "C" standing for "clear") provide only unpolarized data. The PAMs' unpolarized total brightness is reconstructed from sets of polarized observations, while the CAMs are produced from unpolarized observations which are made at a reduced cadence. So while the total brightness layer of the PAMs is conceptually equivalent to the CAMs, PAMs average together more data, which is expected to provide better signal-to-noise. But for memory- or bandwidth-constrained situations that don't require polarization, CAMs provide largely equivalent data at one-third the file size.
```

Here are some possible additions to this paragraph. I think two product codes aren't enough to be alphabet soup, and I think their differences and similarities are subtle enough to deserve a few more sentences. I think especially the fact that PAMs include "clear" data, and the differences between that and the CAMs, deserves careful attention and emphasis.

jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #929, opened 2026-05-05
Set DB polarization for L3 *IM, *TM

svank:
We weren't setting this field, meaning e.g. in the files dashboard you won't see them if you filter to clear polarization states.


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #932, opened 2026-05-07
Properly interpolate starfield models and use correct time windows

### LINKED ISSUE #871: Sort out starfield timing problems
Right now the starfield model is valid for too long of a time in the pipeline. It results in weird holes.

### LINKED ISSUE #872: Interpolate between starfield models
We want to avoid the weird jumpy behavior we're seeing now

lowderchris:
Updates to starfield subtraction

### DIFF
--- a/changelog/932.feature.rst
+++ b/changelog/932.feature.rst
@@ -1,0 +1,1 @@
+Allows interpolation between starfield models.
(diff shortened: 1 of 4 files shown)


jmbhughes (approved):

jmbhughes:
I think this is good, but I'm tired and might wait to merge until morning so I can see it with fresh eyes.

PR #934, opened 2026-05-08
Outlier check logic was backwards in L3_PAM/L3_CAM

jmbhughes:
This corrects the logic.


svank (approved):
I think this only affects the logic that ensures outliers get marked as processed, even though they're filtered out from being passed into the flow. The only consequence would be that that unmarked outliers get seen and rejected again and again every time the scheduler runs.

PR #940, opened 2026-05-12
Make alignment faster

svank:
This makes the alignment code run in about 1/3 the time at the same core count (and I think we can use a few more cores---it's configurable now!).

For a typical image, we had been runing 175 iterations, each finding a best-fit WCS for a small subset of catalog stars, and we used the single WCS which reported the lowest RMS error for its subset of stars. Across all iterations, the spread of fitted CRVALs looks like this:
image
The color is the reported RMS error for that WCS's subset, and it noticeably doesn't really trend with distance to the center of the CRVAL cloud. If we assume the cloud center is the true answer, I bet these residuals are dominated by error in the identified star centroids, and so there's nothing special about the one fit with the lowest residual. This PR changes to using the median CRVAL and CROTA across the iterations.

With that in place, we can run a lot fewer iterations. We can compare the median WCS we get after 50 iterations to what we get after 175 iterations, by taking a set of pixels at the image edges and going pixel --wcs1-> world --wcs2-> pixel and computing the RMS difference i pixels. That distribution, over 20,000 WFI images, looks like this:
image

So 50 iterations looks basically as good as 175. That's the bulk of the speedup here.

This also reduces import overhead for the parallel workers. For one worker, startup time goes from ~4s to 2.5s.

### DIFF
--- a/changelog/940.feature.rst
+++ b/changelog/940.feature.rst
@@ -1,0 +1,1 @@
+Alignment uses a median of the iterations that are run, allowing much fewer iterations to be used.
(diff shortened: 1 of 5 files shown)


svank:
And here's a sanity check of old vs. new for one image
image

jmbhughes (approved):

PR #944, opened 2026-05-13
Use a BlockMasker for remove_starfield

svank:
On `remove_starfield` `main`, we can now be clever about not reprojecting the empty parts of our input trefoils.

The output frame is divided into 128x128 cells. Each cell's center is transformed to the input image's frame, and a 128x128 cell around that transformed center is sampled. If the entire cell is zero or nan, the corresponding cell in the output frame is marked "bad" and not reprojected into. Contiguous "good" cells are merged into strips, and each strip is reprojected into as a separate reproject call. For one input image, those strips overlaid on the output starmap, look like this:
image

Trefoils fill about 2/3 of the image frame, so this means we get through the reprojection part of starfield generation in about 2/3 the time. Including the reduction time, starfield generation time for one test went from 25 minutes to 17.5 minutes.

Since we only transform each cell's center into the output frame, not its size or orientation, we sometimes erroneously clip the very edges of input images, but I think the speed-up is worth it. We can afford to use many more input images, so we're still netting many more pixels.


lowderchris:
A note - tests are failing from not running that updated version of remove_starfield

lowderchris (approved):

jmbhughes:
I'll merge this when we update the remove-starfield dependency in the requirements.txt for punch-mission and in the operational environment.

PR #953, opened 2026-05-19
Change PFW order for NFI

### LINKED ISSUE #807: Check the NFI filter wheel orientation
We should use the F-corona to compare between WFI and NFI to make sure the data is oriented properly. There's concern we may have M and P polarizations flipped.

svank:
Per discussion with Glenn, the filters in the NFI filter wheel are flipped to account for the fact that the wheel was mounted backwards. Here are two places where that should be changed---are there others?

### DIFF
--- a/changelog/953.bugfix.rst
+++ b/changelog/953.bugfix.rst
@@ -1,0 +1,1 @@
+Change PFW order for NFI because it was correct without our flip.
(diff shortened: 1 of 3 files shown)


s0larish (approved):

jmbhughes:
@s0larish does your approval mean that you've tested this flipped order and it matches WFI?

s0larish:
The hallway chat is in agreement with the color flip we saw in NFI today morning. This change seems to correct the flip that was seen in NFI as compared to WFI images.

jmbhughes:
pre-commit.ci autofix

d-lamb:
Closes #807.

PR #955, opened 2026-05-19
Change handling of celestial coordinates

svank:
* This changes how we convert between HP and celestial WCSes, to use Astropy-native HP <-> GCRS conversions, rather than our home-brew method. This conversion is specifically for CRVAL. I *think* the reason I wrote my own is because Astropy doesn't want to do the conversion without the radial dimension set, since it routes HP through heliocentric and HCRS to get to GCRS, so there's a change in origin. The home-brew just did the rotations directly, but it assumed the observer was at Earth, and so by registering the transformation with Astropy, we broke ICRS <-> HP for non-Earth-based observers (because it enabled a ICRS -> GCRS -> HP route, which is shorter). So, for example, when I was doing WISPR coordinates with punchbowl imported, my ICRS <-> HP conversions were broken.
  * This PR just adds a radial distance of 9e99 lightyears and uses the Astropy-native transformations.
  * The HP -> GCRS function is kept, but no longer registered in the Astropy transformation graph, because I didn't want to mess with the function that computes the rotation angle between HP and GCRS, which we use to set CROTA in the converted WCSes.
 * I noticed that, both before and after these changes, CROTA isn't round-tripping from HP -> GCRS -> HP, or GCRS -> HP -> GCRS. It's off by 0.025 degrees, which makes an RMS error of 0.4 px across the image. That's good enough for the HP side, I think, so to avoid having to figure out where this error comes from, `punch_io` now reads and writes the celestial WCS directly, rather than treating HP as the main one and doing conversions back and forth as needed.
   * We now have an `NDCube` subclass that holds the celestial WCS. There are various spots where we make new cubes with e.g. the same data and wcs but a new Meta, and rather than updating all those to copy through the second WCS, I made a utility function to simplify the operation and make sure everything gets copied through.


jmbhughes:
Thank you @svank ! I shall begin my review.

punchbowl/level1/tests/test_stray_light.py
```diff
@@ -45,7 +45,7 @@ def test_estimate_stray_light_runs(tmpdir, sample_ndcube):
         paths.append(path)
 
     with disable_run_logger():
-        cube = estimate_stray_light.fn(paths, 3, num_loaders=2, num_workers=2)
```
jmbhughes, line 48: Was dropping `num_loaders` intentional?

punchbowl/levelq/tests/test_pca.py
```diff
@@ -19,13 +19,16 @@ def test_find_bodies_in_image_quarters():
     wcs.wcs.crpix = 1024.5, 1024.5
     wcs.wcs.cdelt = 0.008333333333, 0.008333333333
     wcs.wcs.dateobs = '2025-07-05T12:56:21.706'
+    wcs.wcs.aux.hgln_obs = 0
+    wcs.wcs.aux.hglt_obs = 7.19
+    wcs.wcs.aux.dsun_obs = 150000000000
```
jmbhughes, line 24: This is the first place I've seen these set. Can you elaborate? Should we be doing this elsewhere?

scripts/create_test_data.py
```diff
@@ -63,7 +62,7 @@ def create_f_corona_test_data(path="../punchbowl/level3/tests/data/"):
 #     data = np.ones((2, 4096, 4096), dtype=np.float32)
 #     uncertainty = StdDevUncertainty(np.sqrt(np.abs(data)).astype(np.float32))
 #
-#     d = NDCube(data=data, uncertainty=uncertainty, wcs=WCS(h), meta=m)
+#     d = PUNCHCube(data=data, uncertainty=uncertainty, wcs=WCS(h), meta=m)
 #     file_path = os.path.join(path, "test_header_validation.fits")
 #     write_ndcube_to_fits(d, file_path, overwrite=True)
```
jmbhughes: Can we just go ahead and delete this commented out code?

punchbowl/data/punch_io.py
```diff
@@ -280,14 +281,14 @@ def write_quicklook_to_mp4(files: list[str],
     subprocess.run(ffmpeg_command, check=False)  # noqa: S603
 
 
-def write_ndcube_to_fits(cube: NDCube,
+def write_ndcube_to_fits(cube: PUNCHCube,
```
jmbhughes, line 284: Should we rename this to `write_punchcube_to_fits`? I know it would break a lot of example code and trigger another large commit, but it seems more precise.

punchbowl/data/punch_io.py
```diff
@@ -415,6 +417,17 @@ def load_ndcube_from_fits(path: str | Path, key: str = " ", include_provenance:
...
+        if key == " ":
+            try:
+                celestial_wcs = WCS(header, hdul, key="A")
+                # If we're loading *lots* of cubes at once, keeping two copies of the distortion tables can matter.
+                # Since they're identical, let's de-duplicate them.
+                celestial_wcs.cpdis1 = wcs.cpdis1
+                celestial_wcs.cpdis2 = wcs.cpdis2
+            except KeyError:
+                # Raised if there isn't a WCS under the "A" key
+                pass
```
jmbhughes: We should at least warn that there's something funky happening if the celestial wcs doesn't exist.

jmbhughes (commented):
Thanks for this change. I had a few comments, but I think it's pretty much ready.

punchbowl/data/punch_io.py
jmbhughes: Same for the loading function.

punchbowl/level1/tests/test_stray_light.py
svank: This was to get the test to pass---at some point we removed `num_loaders` from `estimate_stray_light`

punchbowl/levelq/tests/test_pca.py
svank: After removing the custom GCRS <-> HP conversion, routing the transformation through Astropy requires these to be set since those frames need observer locations to transform. (Our custom routine just assumed the observer locations were the same between the two frames.) These numbers are numbers I pulled from a real FITS file (and then rounded).

scripts/create_test_data.py
svank: Oh shoot, yes

punchbowl/data/punch_io.py
svank: It would break a ton of my notebooks too, especially the loader. Strictly speaking, `isinstance(load_ndcube_from_fits(...), NDCube)` is True...

I don't like the inconsistency but I also don't like renaming the functions either.

svank: Since a PUNCHCube is 99.9% just an NDCube, I feel like there's something to be said for using the term an end user may already be familiar with, rather than hiding the NDCube-ness behind our custom name.

jmbhughes (approved):

PR #956, opened 2026-05-20
Remove unused DB column

svank:
We briefly used a Priority column to influence image formation attempts, before switching to a least-recently-retried system. We can get rid of the column definition, and then remove it from the production DB as well.

(190 didn't have the column in the DB, so TLM ingest was failing.)


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #960, opened 2026-05-22
Adds functionality to label satellites when plotting

jmbhughes:
The aurora community was trying to identify which satellite was which when plotting, so I added a quick option to the plotter.

### DIFF
--- a/punchbowl/data/visualize.py
+++ b/punchbowl/data/visualize.py
@@ -206,6 +206,7 @@ def plot_punch(  # noqa: C901
     grid_spacing: int = 15,
     grid_alpha: float = 0.25,
     title_prefix: str | None = None,
+    label_satellites: bool = False,
     colorbar: bool = True,
     colorbar_label: str = "Mean Solar Brightness (MSB)",
     persistence_array: np.ndarray | PUNCHCube | None = None,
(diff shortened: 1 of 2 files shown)


punchbowl/data/visualize.py
```diff
@@ -339,6 +342,20 @@ def plot_punch(  # noqa: C901
...
+                             "CTRXWFI2", "CTRYWFI2",
+                             "CTRXWFI3"," CTRYWFI3",
+                             "CTRXNFI4","CTRYNFI4"]
+        for i in range(0, 8, 2):
+            satellite = position_keywords[i][-4:]
+            if position_keywords[i] in cube.meta and position_keywords[i+1] in cube.meta:
+                x, y = cube.meta[position_keywords[i]].value, cube.meta[position_keywords[i+1]].value
+                if x != -1 and y != -1:
+                    ax.text(x, y, satellite, color="white",
+                            verticalalignment="bottom", horizontalalignment="left",
```
svank: Center alignment maybe makes more sense here.

svank (changes requested):

punchbowl/data/visualize.py
jmbhughes: True. This is why we review :)

svank (approved):

PR #964, opened 2026-06-01
Better spacecraft position handling

svank:
* When we write L0s, we:
  * use the GPS position closest to the middle of the exposure
  * use DATE-OBS, not DATE-BEG, to convert between frames. (For heliocentric frames, the difference is 1/2 exposure time * Earth's orbital motion, which comes out to ~0.1 R_earth.)
  * write position and velocity in several additional frames:
    * GEO: geo-centric (ITRS) cartesian x/y/z, to pair with vx/vy/vz
    * GEI: geocentric earth equatorial
    * GSE: geocentric solar ecliptic
    * There's at least one other instrument writing each of GEI and GSE, and the three-letter codes are on https://docs.sunpy.org/en/stable/reference/coordinates/index.html, so they must be at least quasi-standard. GEO is the raw GPS coordinate converted to cartesian, so it's a natural choice for providing minimally-processed values.
* When we read a FITS file, we now read the celestial WCS into a GCRSWCS object, which subclasses WCS to accept & return coordinates in GCRS. It reads the GEO* keys from the header to get the coordinate origin position and velocity. Using this simplifies the alignment code.
* `NormalizedMetadata.get` now returns an actual value, not a MetaField, because otherwise the ability to provide a default value doesn't make sense.


jmbhughes (approved):
@svank I wonder if we should change the quaternion to WCS conversion to use the middle XACT packet instead of interpolating the before and after ones. That can be a separate PR though.

PR #966, opened 2026-06-02
Allows per-polarizer state quartic coefficient calibration files

lowderchris:
Updates to allow per-polarizer state quartic coefficient calibration files with updates values from James Tappin.

TODO
- [x] Confirm strategy
- [ ] Update calibration files accordingly


punchbowl/auto/flows/level1.py
```diff
@@ -412,7 +414,7 @@ def get_quartic_model_paths(level0_files, pipeline_config: dict, session=None):
 
 def get_quartic_model_path(level0_file, pipeline_config: dict, session=None, reference_time=None):
     best_model = (session.query(File)
-                  .filter(File.file_type == "FQ")
+                  .filter(File.file_type == f"F{level0_file.polarization}")
```
jmbhughes: Can you check that the level 0 file polarization is always set as R, M, Z, or P in the database? I want to avoid problems when we deploy.

jmbhughes (commented):
I think you got everything. I just want to double check this one detail commented below. Thanks!

punchbowl/auto/flows/level1.py
jmbhughes: Sam checked. It isn't true. We should instead use the second character of the file type for the level 0 file.

PR #967, opened 2026-06-02
De-Prefect core code

svank:
Here's a nice-to-have, which I thought would help me right now but didn't. (I'm trying to run level1_early in parallel from a notebook and it kept crashing. I thought it was a "Prefect threads while forking" problem, but it turned out to be a "setting thread limits while forking" problem.)

All calls to `get_run_logger` are replaced with a utility function that checks if we're running under Prefect, and returns the Prefect logger if we're running under Prefect, and a plain Python logger otherwise.

`@punch_task` and `@punch_flow` also check if we're running under Prefect, and if not, just return the function. (They do add a `.fn` attribute to the function that just points to the function itself, so that API still works.)

We don't check for "debug mode" if we're not running under Prefect.

I don't know if this fully removes all Prefect everything from the stuff that was never `punchpipe`, but it moves us in that direction. This will help with setting up more minimal dependency options, and will simplify things for community members who want to run parts of our code in their own notebooks.


punchbowl/auto/flows/dynamic_wfi_stray_light.py
```diff
@@ -61,7 +61,7 @@ def construct_dynamic_stray_light_check_for_inputs(session,
                                                    pipeline_config: dict,
                                                    reference_time: datetime,
                                                    reference_file: File):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 65: I think you need to import the new function in this file still.

punchbowl/auto/flows/fcorona.py
```diff
@@ -18,7 +18,7 @@
 
 def f_corona_background_query_ready_files(session, pipeline_config: dict, reference_time: datetime,
                                           reference_file: File):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 22: I think you need to import the new function in this file still.

punchbowl/auto/flows/level0.py
```diff
@@ -364,7 +364,7 @@ def ingest_tlm_file(path: str,
 
 @task
 def unpack_n_bit_values(packed: bytes, byteorder: str, n_bits=19) -> np.ndarray:
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 368: I think you need to import the new function in this file still.

punchbowl/auto/flows/level2.py
```diff
@@ -30,7 +30,7 @@ def level2_query_ready_clear_files(session, pipeline_config: dict, reference_tim
 
 
 def _level2_query_ready_files(session, polarized: bool, pipeline_config: dict, max_n=9e99):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 34: I think you need to import the new function in this file still.

punchbowl/auto/flows/level3.py
```diff
@@ -99,7 +99,7 @@ def get_fcorona_pairs(session, files: list[File], model_type="PF"):
 
 @task(cache_policy=NO_CACHE)
 def level3_PTM_query_ready_files(session, pipeline_config: dict, reference_time=None, max_n=9e99):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 103: I think you need to import the new function in this file still.

punchbowl/auto/flows/simulate.py
```diff
@@ -68,7 +68,7 @@ def simpunch_core_flow(
         transient_probability: float = 0.03,
         shift_pointing: bool = False) -> list[str]:
 
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 72: I think you need to import the new function in this file still.

punchbowl/auto/flows/levelq.py
```diff
@@ -25,7 +25,7 @@
 
 @task(cache_policy=NO_CACHE)
 def levelq_QNN_query_ready_files(session, pipeline_config: dict, reference_time=None, max_n=9e99):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 29: I think you need to import the new function in this file still.

punchbowl/auto/flows/stray_light.py
```diff
@@ -22,7 +22,7 @@ def construct_stray_light_check_for_inputs(session,
                                            pipeline_config: dict,
                                            reference_time: datetime,
                                            reference_files: list[File]):
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 26: I think you need to import the new function in this file still.

punchbowl/auto/flows/velocity.py
```diff
@@ -17,7 +17,7 @@
 @task
 def level3_vam_query_ready_files(session, pipeline_config: dict, reference_time: datetime=None, max_n: float=100):
     flow_type = "L3_VAM"
-    logger = get_run_logger()
+    logger = get_logger()
```
jmbhughes, line 21: I think you need to import the new function in this file still.

punchbowl/prefect.py
```diff
@@ -26,19 +29,60 @@ def completion_debugger(task: Task, task_run: TaskRun, state: State) -> None:
...
+def _compatability_decorator(func: Callable) -> Callable:
+    """Make wrapped functions have a .fn attribute like Prefect Flows and Tasks."""
+    func.fn = func
+    return func
+
+
+@cache
+def detect_if_running_in_prefect() -> bool:
+    """Determine if we're running under Prefect."""
+    return runtime.flow_run.name is not None
```
jmbhughes, line 87: Is there a better way to check if Prefect is running? This is an interesting approach that may be the best, but I just wanted to ask.

jmbhughes (changes requested):
I think this is a good idea, but there are places where you don't import the new `get_logger` and the old `get_run_logger` Prefect import is still present.

I also see some tests failed because `FAILED punchbowl/level2/tests/test_resample.py::test_reproject_many_flow - AttributeError: 'function' object has no attribute 'submit'`. We'll need a work around for that too. 

And finally we'll need a changelog.

punchbowl/prefect.py
svank: It's what I could find online

svank:
OK, the imports are fixed. I was counting on Ruff to catch that, but I don't think it runs inside `auto/`.

I'll have to think about `submit()` later. Maybe we add it as a wrapper that just calls the function and returns an object that has the return value under `.result()`.

jmbhughes:
> OK, the imports are fixed. I was counting on Ruff to catch that, but I don't think it runs inside `auto/`.

`auto` offended Ruff so much I turned it off until we could find time to clean it up.

svank:
OK, this PR has support for `.submit()` now.

jmbhughes (approved):

PR #973, opened 2026-06-08
Alignment changes

svank:
Here's a batch of miscellaneous alignment changes, including:

* Simplify the top-level align_task
* Use GCRS coordinates when we do `find_catalog_in_image`---otherwise, it's possible to have a star that right at the edge of the image in ICRS, but is just outside the image in GCRS. This changes the typical flow in that `prep_star_coords` has to be called before `find_catalog_in_image`, not after, and so I changed some input and return types to simplify that.
* Fix for RA wrap point---the outer code that does the median of all the iterations was assuming RA would fall between 0 and 360, but the inner code didn't do any wrapping, so it could produce RA values just below 0 or just above 360. In a notebook case where an L1 alignment result was being fed back into the alignment code, astrometry chocked because the RA hint was above 360.


svank:
pre-commit.ci autofix

jmbhughes:
pre-commit.ci autofix

jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):

PR #974, opened 2026-06-08
Alignment speedup

svank:
Big alignment speedup, by:

* Not handling frames or units in our inner-most loop---this is a many-times speedup, while using fewer cores (at 50 iterations, the cost of launching even one worker isn't worth the parallelism)
* Forward-applying the distortion map once to the identified star locations, instead of repeatedly backwards-applying the map to expected star locations. `world_to_pixel` is significantly slower with a distortion map since going backwards is an iterative process, but it's not the main bottleneck, so this gets us maybe a 10-20% speed-up.


svank:
pre-commit.ci autofix

punchbowl/level1/alignment_parallel.py
```diff
@@ -106,13 +110,14 @@ def _residual(params: Parameters,
     guess_wcs.cpdis1 = guess_wcs.cpdis1
```
courtneypeck: Is there a reason this is getting set equal to itself?

punchbowl/level1/alignment.py
```diff
@@ -445,9 +445,13 @@ def mask(observed: np.ndarray) -> np.ndarray:
     guess_wcs.wcs.crpix = image_center
     guess_wcs.wcs.cdelt = image_wcs.wcs.cdelt
     guess_wcs.sip = None
+
     if distortion is not None:
-        guess_wcs.cpdis1 = distortion.cpdis1
-        guess_wcs.cpdis2 = distortion.cpdis2
+        for i in range(len(observed)):
```
courtneypeck, line 461: It may be worth testing if using list comprehension instead of the for loop could speed things up:

`observed_distorted = np.array([[i+distortion.cpdis1.get_offset(i, j),j+distortion.cpdis1.get_offset(i,j)] for i,j in observed])`

Caveat: At this time, I don't have the PUNCH repo fully configured, data access, or background knowledge of the code to test/know if this suggested code snippet works, so I'm going off of what I can gather from the PR and the astropy documentation about the array shapes/types.

punchbowl/level1/alignment.py
```diff
@@ -464,12 +468,16 @@ def mask(observed: np.ndarray) -> np.ndarray:
     rng = np.random.default_rng(seed=1)
     results = []
     observed_tree = KDTree(observed)
-    mp_context = multiprocessing.get_context("forkserver")
-    with ProcessPoolExecutor(n_workers, mp_context) as p:
-        for _ in range(n_rounds):
-            sample = catalog_stars[rng.choice(indices, 15, replace=False)]
-            results.append(p.submit(refine_pointing_single_step, guess_wcs, observed_tree, sample, fix_pv=True))
-    results = [w.result() for w in results]
+    star_samples = [catalog_stars[rng.choice(indices, 15, replace=False)] for _ in range(n_rounds)]
```
courtneypeck: What is the significance of the number 15? It looks like it's dictating the size of the output star samples array. Is the code sensitive at all to changing this number? Does it ever need to be adjusted? 

I see that this value was set prior to your PR, so these questions are really out of curiosity/clarification.

courtneypeck (commented):
I'm obviously still learning the codebase, so I've noted a few things that stuck out to me that may or may not be necessary to change.

jmbhughes:
I don't yet understand why the tests are failing here. They pass locally. It's some interplay between Prefect and GitHub Actions.

jmbhughes (approved):

PR #975, opened 2026-06-08
Do reprojection with 32-bit data

svank:
This is a 15% speedup to the entire `level2_clear` flow when running 50 simultaneous workers. We've decided in the past that 32 bits is fine since we compress the output data and don't expect meaningful signal anyway in those last bits.

If you sample the CPU instructions-per-cycle statistic while running L2, you can see that the values are higher after this change, supporting the interpretation that this is a memory bandwidth thing (more instructions run per clock cycle because of less waiting for data).


svank:
pre-commit.ci autofix

lowderchris (approved):
Looks good. The CI was failing because it did not have the new release of reproject. Re-running that now just to confirm tests passing locally.

lowderchris:
@jmbhughes - we should make sure to update reproject on the production / test servers before running with this.

jmbhughes (approved):

PR #978, opened 2026-06-09
Daily quicklook scheduling

### LINKED ISSUE #972: Daily movies need more care in scheduling
On June 5, no WFI 1 or 3 movies had been made since May 18.

If downlinks are delayed, so that there's nothing new on the day in question, it may be that the daily movie generation doesn't see any new movies for "today" and so it doesn't generate anything.

lowderchris:
Updates daily movie quicklook scheduling to include files created in the past 24 hours, not by the date of observations. This was leading to movies not being generated, presumably from processing delays.

### DIFF
--- a/changelog/978.bugfix.rst
+++ b/changelog/978.bugfix.rst
@@ -1,0 +1,1 @@
+Updates daily movie quicklook scheduling to include files created in the past 24 hours, not by the date of observation.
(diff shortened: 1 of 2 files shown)


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):
Also if we're reprocessing old files, we wouldn't be making movies under the old code.

PR #979, opened 2026-06-11
Remove space in WFI3 plotting keywords preventing it from plotting

jmbhughes:
There was a bug in the WFI3 label that included a space. It then didn't find the keyword in the metadata and thus didn't plot the label as requested.


courtneypeck (approved):
Beautiful. Amazing. Well done.

PR #983, opened 2026-06-14
Remove old quadprog approach

jmbhughes:
We're not using the quadratic programming approach to F-corona modeling, so I'm removing it and removing the dependency.


jmbhughes:
pre-commit.ci autofix

PR #986, opened 2026-06-15
Enhance "getting help" in the docs

svank:
To make the Help page more of a one-stop shop, I added the SOC email address. I also re-arranged it a bit in a way I think flows better.

I also linked the Help page from the main landing page, so it's more prominent.

My goal is to be able to put the URL on a slide and have that provide a more direct and complete route for getting help.


svank:
pre-commit.ci autofix

jmbhughes (approved):

PR #992, opened 2026-06-18
Update pyproject.toml to support new developer's environment

courtneypeck:
This PR is closely related to https://github.com/punch-mission/punch-mission/pull/47 but neither explicitly depends on the other. 

This PR reorganizes the punchbowl pyproject.toml to facilitate the PUNCH lite user install with a specific emphasis on removing the core dependency of astrometry. To do this, I needed to adapt the punchbowl pyproject.toml since this file was using astrometry as a required dependency. 

@jmbhughes I am only adding you as a reviewer at this moment, though it would be helpful if everyone on the team could test these two related PRs before merging.


courtneypeck:
Nice. All jobs failed!

pyproject.toml
```diff
@@ -6,32 +6,20 @@ build-backend = "setuptools.build_meta"
...
-    "sunpy[all]",
-    "pandas",
-    "ndcube",
-    "matplotlib<=3.10.9",
-    "ccsdspy",
-    "prefect[dask]",
-    "regularizepsf",
-    "solpolpy",
-    "scipy",
+    "astroscrappy",
```
jmbhughes:
```suggestion

```
Not used anymore

pyproject.toml
```diff
@@ -6,32 +6,20 @@ build-backend = "setuptools.build_meta"
...
-    "setuptools",
-    "reproject",
-    "pylibjpeg",
+    "matplotlib<=3.10.9",
+    "ndcube",
+    "numpy",
     "python-dateutil",
-    "remove_starfield>=0.1",
-    "pylibjpeg[openjpeg]",
+    "quadprog",
```
jmbhughes:
```suggestion

```
Not used anymore

jmbhughes (commented):
Thanks!

lowderchris (commented):
I think these changes will make the install process much easier for folks, so thank you! My only question - it seems like it's no longer installing ruff as part of the GitHub CI build, and causing things to fail. But I do see in the ci.yaml that it's calling pip install ".[dev]" 🤔

courtneypeck:
Woohoo! Most checks are passing after much effort! @jmbhughes Do we expect the changelog: absent check to fail? How do I fix this?

jmbhughes:
> Do we expect the changelog: absent check to fail? How do I fix this?

@courtneypeck You can add a changelog entry following the README in that folder https://github.com/punch-mission/punchbowl/tree/main/changelog. If you need help, let me know.

courtneypeck:
Thanks @jmbhughes. The check is successful now.

svank (approved):

courtneypeck:
After discussion with Marcus, we decided that punchbowl should hold the pyproject.toml environment file rather than punch-mission. I have therefore merged the pyproject.toml from my punch-mission branch to this branch (and deleted the files in punch-mission that are now no longer necessary). This should make it easier for us to manage dependencies across punchbowl and our other repositories. 

If you'd like to test creating the environment using the updated pyproject.toml on this branch in punchbowl, that would be great. Otherwise, unit testing and CI checks are passing for me (with the exception of 4 unit tests that are failing for openjpeg reasons) when I create the environment in punchbowl, so it should be good to go based on all the feedback from the team.

PR #1053, opened 2026-07-10
Avoid overflows when merging at 32 bits

svank:
In #975 we got a speed boost by running our L1 -> L2 reprojection at 32 bits. That turned out to have the side effect of causing holes in our L2s wherever the uncertainty was inf.

Inf uncertainty occurs where the data values are fill values---e.g. where there were saturated values or cosmic rays. In the merging code, those fill values have their uncertainty changed from inf to `very_large`, so that the fill values get used if there's no other option, but at regions of overlap between WFIs, the fill values get a weight that's infinitesimally small and the other WFI's value gets used instead.

When we changed to 32 bits, the `very_large` value was too large, so (especially when the weights are computed as `1 / uncertainty ** 2`) we overflowed back to inf. This new value is still extremely large but doesn't overflow.


jmbhughes (approved):
Seems good to me. Let's merge after @coronakenny has a chance to comment.

coronakenny (approved):
Looks good to me, too!

svank:
pre-commit.ci autofix

PR #1060, opened 2026-07-14
Coalign MZP before polarization resolution

### LINKED ISSUE #1058: Implement coalignment in L1 frame before polarization resolution and mosaic assembly
Since the spacecraft pointing can drift slightly through an imaging sequence, the stars may not be precisely aligned across the MZP triplet. We should reproject into a common frame before we convert from instrument MZP to solar MZP.

(solpolpy currently assumes the instrument MZP directions are uniform across the image, i.e. that instrument MZP images are provided in the instrument frame, so we can't just wait until the mosaic frame to do the resolution.)

svank:
Closes #1058

Now we co-align MZP triplets before converting the polarization. This is by reprojecting the three images into the same target frame. That frame is the Z frame, but with the distortion maps removed (to reduce the workload for computing both this reprojection and the following L1 -> L2 reprojection). Reprojection always adds a bit of blurriness, so the Z image has to be reprojected into its own frame so it gets the same blurring as M and P. The common L1 frame is also scaled up by 40%, which reduces the amount of extra blurring we get by adding this round of reprojection.

This extra reprojection adds a compute speed penalty, but that's partially mitigated by the fact that our co-aligned L1s are in the exact same frame, so when we do the common-L1 -> L2 reprojection, we can stack the three images and do one reproject call, and the coordinate transformations only have to be computed once (rather than separately for the M, Z and P images).

Here are two before-and-after comparisons. Each is a cutout of an L2 image. The columns are tB, pB, and pB', and the rows are before this PR, with this PR, and the difference. In pB and especially in pB', there are bipoles and negative values wherever there are bright stars in the images, and these unphysical artifacts are reduced with this PR. The tB difference shows the extra blurring due to the extra reprojection (and the difference frames' colorbars are zoomed in 10x), but that blurring isn't visually apparent in the tB panels themselves.

image
image

### DIFF
--- a/changelog/1060.feature.rst
+++ b/changelog/1060.feature.rst
@@ -1,0 +1,2 @@
+In L2 mosaic assembly, polarized inputs are now co-aligned before the polarization is converted to solar MZP, to
+account for small pointing drift through the imaging sequence.
--- a/changelog/1060.feature.rst
+++ b/changelog/1060.feature.rst
@@ -1,2 +1,2 @@
-In L2 mosaic assembly, polarized inputs are now co-aligned before the polarization is converted to solar MZP, to
+In L2 mosaic assembly, polarized inputs are now co-aligned before the polarization is converted to MZP_solar, to
 account for small pointing drift through the imaging sequence.
--- a/punchbowl/level2/flow.py
+++ b/punchbowl/level2/flow.py
@@ -13,7 +13,7 @@ from punchbowl.level2.finalize import finalize_output
 from punchbowl.level2.merge import merge_many_clear_task, merge_many_polarized_task
 from punchbowl.level2.polarization import resolve_polarization_task
 from punchbowl.level2.preprocess import preprocess_trefoil_inputs
-from punchbowl.level2.resample import find_central_pixel, reproject_many_flow
+from punchbowl.level2.resample import coalign_L1_mzp, find_central_pixel, reproject_many_flow
 from punchbowl.prefect import get_logger, punch_flow
 from punchbowl.util import load_image_task, output_image_task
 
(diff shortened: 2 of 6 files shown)


punchbowl/level2/flow.py
```diff
@@ -111,25 +111,80 @@ def level2_core_flow(data_list: list[str] | list[PUNCHCube], # noqa: C901
...
+            coaligned_ordered_cubes = [None] * len(ordered_data_list)
+            for i in range(0, len(POLARIZED_FILE_ORDER), 3):
+                cubes = ordered_data_list[i:i + 3]
+                if any(c is None for c in cubes):
+                    continue
+                # We reproject the images into the Z frame. The Z image must be reprojected as well, so everyone gets
+                # the same amount of anti-aliasing blur. We reproject into a larger (scaled up) pixel grid,
+                # to minimize the impact this extra round of reprojection has on the blur of the final images. The
+                # co-aligned frame also strips out the distortion table, since that's just extra coordinates work we
+                # don't need.
```
jmbhughes, line 131: Why don't we need the distortion?

jmbhughes (approved):

punchbowl/level2/flow.py
svank: As we're reprojecting the data, we're reprojecting from the actual, distorted frames, but it doesn't matter whether we're reprojecting into a frame that has the same distortions, or an ideal frame without distortion---either is fine for converting the polarization. The ideal frame will give us faster reprojections, since the distortion maps don't have to be applied to the target frame (nor to the source frame when we do the coaligned-L1 to L2 reprojection)

jmbhughes: That makes sense. Thanks.

changelog/1060.feature.rst
```diff
@@ -0,0 +1,2 @@
+In L2 mosaic assembly, polarized inputs are now co-aligned before the polarization is converted to solar MZP, to
```
s0larish: shall we put mzpsolar to keep it consistent with solpolpy naming?

PR #1062, opened 2026-07-14
Script updates

svank:
* Better error reporting in `import_export`
* `productionify` supports running to generate jp2s and/or .sha files without changing the version string.

### DIFF
--- a/changelog/1062.feature.2.rst
+++ b/changelog/1062.feature.2.rst
@@ -1,0 +1,1 @@
+Better error reporting in import_export.py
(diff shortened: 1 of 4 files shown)


jmbhughes (approved):
Let's have @Rokkaan5 review but I think it's good.

Rokkaan5 (commented):
I'm not as well versed in what this particular change is to give any specific feedback, but I have looked over the file changes and nothing particularly stood out to me as concerning. (Not sure how to "test" this change either, if applicable.) 
Honestly more relying on Marcus' acknowledged approval of this, but I also think it's okay. :)

PR #1075, opened 2026-07-18
Switch to UV for CI

### LINKED ISSUE #984: Switch to uv for CI runs
We could speed up the CI by using uv to install the packages as well as potentially caching the environment (assuming it doesn't change).

jmbhughes:
## PR summary

This PR switches to UV and closes #984 . In doing so it switches to uv for CI actions and adds a concurrency limit and timeout for GitHub Actions.

## Todos

- [x] make sure it runs properly
- [x] fix whatever is wrong with 3.14

## Test plan

I think we just have to let GitHub do its thing. 

## Breaking changes

None

## Related issues

#984 

## AI usage

None

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,10 +35,10 @@ jobs:
     - name: Install dependencies
       run: |
         uv sync --no-sources --all-extras
-        source .venv/bin/activate 
+        source .venv/bin/activate
     - name: Lint with ruff
       run: |
-        ruff check .
+        uv run ruff check .
     - name: Run codespell on source code
       uses: codespell-project/actions-codespell@v2
       with:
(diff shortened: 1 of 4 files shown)


jmbhughes:
pre-commit.ci autofix

jmbhughes:
If this looks good, I can do a similar thing for solpolpy and regularizepsf.

lowderchris (approved):
```
Resolved 362 packages in 1ms
Prepared 348 packages in 10.63s
Installed 348 packages in 699ms
```

Incredible. The only thing that caught my eye reviewing the CI runs on this PR was in the Python setup stage, where it just gave a few messages:

```
Trying to find version for uv in: /home/runner/work/punchbowl/punchbowl/uv.toml
Could not find file: /home/runner/work/punchbowl/punchbowl/uv.toml
Trying to find version for uv in: /home/runner/work/punchbowl/punchbowl/pyproject.toml
Could not determine uv version from uv.toml or pyproject.toml. Falling back to latest.
```

But that might be the intended behavior to not pin a version there and just have it download the latest version.

jmbhughes:
> But that might be the intended behavior to not pin a version there and just have it download the latest version.

We could pin it but doesn't seem strictly necessary.

svank (approved):

jmbhughes:
Note: I learned all uv commands need `--no-sources` if you're intending to use them without the source downloaded even if the environment was set up already that way. It took me a few tries to realize that you need it both when setting up the environment and when running.

PR #1079, opened 2026-07-20
1041 implement flow tracking

### LINKED ISSUE #1041: Implement flow tracking code changes
Implement updated flow tracking approach within punchbowl pipeline code. This involves:
- [x] Reviewing existing code
- [x] Update for modified input data / timing
- [x] Run limited test

lowderchris:
## PR summary

Updates existing flow tracking code to work with existing PUNCH data.

## Todos

- [x] Confirm timing parameters
- [x] Confirm PTM versus CTM
- [ ] Integrate with pipeline config

## Test plan

Outputs will be tested and validated as part of a separate task.

## AI usage

None

### DIFF
--- a/punchbowl/auto/flows/velocity.py
+++ b/punchbowl/auto/flows/velocity.py
@@ -118,7 +118,7 @@ def level3_vam_construct_flow_info(level3_ptm_files: list[File],
             "delta_t": pipeline_config["flows"][flow_type].get("delta_t", 12),
             "sparsity": pipeline_config["flows"][flow_type].get("sparsity", 2),
             "n_ofs": pipeline_config["flows"][flow_type].get("n_ofs", 151),
-            "ycens": pipeline_config["flows"][flow_type].get("ycens", np.arange(30, 90, 10)),
+            "ycens": np.array(pipeline_config["flows"][flow_type].get("ycens", np.arange(30, 90, 10))),
             "rbands": pipeline_config["flows"][flow_type].get("rbands", None),
         },
     )
--- a/punchbowl/auto/flows/velocity.py
+++ b/punchbowl/auto/flows/velocity.py
@@ -2,6 +2,7 @@ import os
 import json
 from datetime import UTC, datetime, timedelta
 
+import numpy as np
 from dateutil.parser import parse as parse_datetime_str
 from prefect import flow, task
 
@@ -114,7 +115,11 @@ def level3_vam_construct_flow_info(level3_ptm_files: list[File],
     call_data = json.dumps(
         {
             "files": [ptm_file.filename() for ptm_file in level3_ptm_files],
-            "reference_time": reference_time.strftime("%Y-%m-%dT%H:%M:%S"),
+            "delta_t": pipeline_config["flows"][flow_type].get("delta_t", 12),
+            "sparsity": pipeline_config["flows"][flow_type].get("sparsity", 2),
+            "n_ofs": pipeline_config["flows"][flow_type].get("n_ofs", 151),
+            "ycens": pipeline_config["flows"][flow_type].get("ycens", np.arange(30, 90, 10)),
+            "rbands": pipeline_config["flows"][flow_type].get("rbands", None),
         },
     )
     return Flow(
--- a/punchbowl/auto/flows/velocity.py
+++ b/punchbowl/auto/flows/velocity.py
@@ -12,7 +12,7 @@ from punchbowl.auto.control.processor import generic_process_flow_logic
 from punchbowl.auto.control.scheduler import generic_scheduler_flow_logic
 from punchbowl.auto.control.util import get_database_session, load_pipeline_configuration
 from punchbowl.auto.flows.util import file_name_to_full_path
-from punchbowl.level3.velocity import track_velocity
+from punchbowl.level3.flow import generate_level3_velocity_flow
 from punchbowl.prefect import get_logger
 
 
@@ -171,7 +171,7 @@ def level3_vam_call_data_processor(call_data: dict, pipeline_config, session=Non
 @flow
 def level3_vam_process_flow(flow_id: int, pipeline_config_path=None, session=None):
     generic_process_flow_logic(flow_id,
-                               track_velocity,
+                               generate_level3_velocity_flow,
                                pipeline_config_path,
                                session=session,
                                call_data_processor=level3_vam_call_data_processor,
--- a/punchbowl/auto/flows/velocity.py
+++ b/punchbowl/auto/flows/velocity.py
@@ -17,7 +17,23 @@ from punchbowl.prefect import get_logger
 
 
 @task
-def level3_vam_query_ready_files(session, pipeline_config: dict, reference_time: datetime=None, max_n: float=100):
+def level3_vam_query_ready_files(session, pipeline_config: dict, reference_time: datetime=None, max_n: float=100) -> list:
+    """Queries files ready for velocity tracking.
+
+    Parameters
+    ----------
+    session
+        Database session
+    pipeline_config : dict
+        Pipeline configuration dictionary
+    max_n : float, optional
+        Max number of files to specify ready, by default 100
+
+    Returns
+    -------
+    list
+        Cleaned ready groups
+    """
     flow_type = "L3_VAM"
     logger = get_logger()
     min_file_count = pipeline_config["flows"]["L3_VAM"]["min_file_count"]
@@ -107,7 +123,27 @@ def level3_vam_construct_flow_info(level3_ptm_files: list[File],
                                    level3_velocity_file: File,
                                    pipeline_config: dict,
                                    reference_time: datetime,
-                                   session=None):
+                                   session=None) -> Flow:
+    """Constructs velocity tracking flow
+
+    Parameters
+    ----------
+    level3_ptm_files : list[File]
+        Input files used for velocity tracking
+    level3_velocity_file : File
+        Output file path for measurement
+    pipeline_config : dict
+        Pipeline configuration settings
+    reference_time : datetime
+        Reference time to use for marking velocity measurement
+    session : _type_, optional
+        Database session, by default None
+
+    Returns
+    -------
+    Flow
+        Velocity tracking flow to run
+    """
     flow_type = "L3_VAM"
     state = "planned"
     creation_time = datetime.now()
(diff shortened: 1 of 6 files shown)


punchbowl/util.py
```diff
@@ -50,7 +50,9 @@ def load_mask_file(path: str) -> np.ndarray:
 
 
 @punch_task
-def output_image_task(data: PUNCHCube, output_filename: str) -> None:
+def output_image_task(data: PUNCHCube,
```
courtneypeck, line 53: Add docstrings.

punchbowl/auto/flows/velocity.py
courtneypeck: Please add docstrings. I know that this is a bit outside scope for this PR since you didn't create this function, but we have to start addressing the technical debt.

punchbowl/level3/flow.py
```diff
@@ -157,17 +157,27 @@ def generate_level3_low_noise_flow(data_list: list[str] | list[PUNCHCube],
 
 
 @punch_flow
-def generate_level3_velocity_flow(data_list: list[str],
+def generate_level3_velocity_flow(files: list[str],
```
courtneypeck, line 160: Add docstrings.

punchbowl/auto/flows/velocity.py
courtneypeck, line 115: Same here. Add docstrings if possible.

courtneypeck (commented):
I'm the docstrings police today :)

punchbowl/auto/flows/velocity.py
svank, line 194: It looks like this isn't actually true. We maybe don't have a meaningful way yet to label VAMs as outliers, but should we be excluding outlier inputs?

svank: Right now, groups of input files will be scheduled after a delay, even if they don't meet the minimum number of files. That made sense for CAMs (which I think this scheduling code is based on), but does it make sense for flow tracking?  Is there a minimum number of input images we actually can't go below?

jmbhughes:
@lowderchris I'm holding my review until the existing notes are addressed.

punchbowl/auto/flows/velocity.py
lowderchris: Ah, yes. We might eventually find a way to flag bad output, but for now if we assume good data in...

There's a bit later in the query where outliers are excluded:

```py
    cleaned_ready_groups = []
    for group in grouped_ready_files:
        group = [f for f in group if f.outlier == 0]
        if group:
            cleaned_ready_groups.append(group)
```

lowderchris: I did some testing taking random sampling of input PTM files, and things broke down below 300 input files. I removed the cutoff time logic to avoid dipping below that, and set the minimum file count in the operational config example below.

lowderchris:
An example of the operational config section for velocity tracking:

```yaml
  L3_VAM:
   run-on: "phoenix"
   enabled: true
   description: "Velocity tracking code"
   min_file_count: 300
   t0: ...
   delta_t: 24
   sparsity: 2
   n_ofs: 151
   ycens: [30, 40, 50, 60, 70, 80, 90]
   rbands: [0, 1, 2, 3, 4, 5]
   tags: ["L3"]
   priority:
     initial: 10003
     seconds: [1]
     escalation: [10003]
   schedule: "0 0 * * *"
   launch_weight: 25
   concurrency_limit: 1
   options:
```

punchbowl/auto/flows/velocity.py
```diff
@@ -106,15 +114,39 @@ def level3_vam_construct_flow_info(level3_ptm_files: list[File],
...
     creation_time = datetime.now()
     priority = pipeline_config["flows"][flow_type]["priority"]["initial"]
     call_data = json.dumps(
         {
             "files": [ptm_file.filename() for ptm_file in level3_ptm_files],
-            "reference_time": reference_time.strftime("%Y-%m-%dT%H:%M:%S"),
+            "delta_t": pipeline_config["flows"][flow_type].get("delta_t", 12),
+            "sparsity": pipeline_config["flows"][flow_type].get("sparsity", 2),
+            "n_ofs": pipeline_config["flows"][flow_type].get("n_ofs", 151),
+            "ycens": np.array(pipeline_config["flows"][flow_type].get("ycens", np.arange(30, 90, 10))),
```
s0larish, line 151: I believe this goes in the scheduler with json file. Can you verify if passing ycens as np array inside dictionary works fine or we should store ycens as a normal list here and the call with np.asarray(ycens) inside track_velocity() later?

punchbowl/level3/velocity.py
```diff
@@ -83,7 +83,7 @@ def preprocess_image(data: PUNCHCube,
     # Replace with appropriate preprocessing needed to clean-up. We need to have finite values for the polar remap
     header = data.meta.to_fits_header(wcs=data.wcs)
 
-    if header["OBS-MODE"] == "Polarized":
+    if header["OBS-MODE"] == "Polar_BpB":
```
s0larish, line 86: As in the past we had shuffled which data product to be used for tracking, it will be helpful to have this in the notes somewhere if product type is changed, then this is the key input point for it.

s0larish (changes requested):

punchbowl/level3/velocity.py
lowderchris: Ah, this bit was just taking the appropriate layer of the data depending on either type of input, polarized or unpolarized. This was just updating to the right key value here.

punchbowl/auto/flows/velocity.py
lowderchris: Ah, this is being stored as a regular list within the config file. The line here just defaults back to that numpy array if values are not found.

jmbhughes:
Ritesh verbally confirmed that he was done with review and Chris addressed the problems.

punchbowl/auto/flows/velocity.py
svank:
```suggestion
                        .filter(File.file_type == "VA")
```
This must have been copied from the PAM flow before a bug was fixed there---this line should query for exiting velocity maps at this point in time

svank: This is removing outlier inputs *after* checking if we exceed the minimum file count. Is the minimum a hard minimum? Should the outlier exclusion happen first? (Maybe it should happen in the database query)

svank, line 54: Since the "progressed" state gets set by the PAM flow, we have to accept them as inputs here, but there's nothing to exclude PTMs that have already been velocity-mapped. That means the number of files returned by this query will grow without bound. We'll probably want something using the `relationships` table and a sub query. I'll suggest something soon

svank: Outlier files are already being removed in `query_ready_files`, so I don't think they need to be excluded here.

punchbowl/level3/velocity.py
```diff
@@ -542,9 +542,9 @@ def track_velocity(files: list[str],
...
         output_meta["DATE-END"] = hdul[1].header["DATE-END"]
 
-    date_beg = datetime.strptime(output_meta["DATE-BEG"].value, "%Y-%m-%dT%H:%M:%S").astimezone()
-    date_end = datetime.strptime(output_meta["DATE-END"].value, "%Y-%m-%dT%H:%M:%S").astimezone()
-    date_avg = (date_beg + (date_end - date_beg) / 2).strftime("%Y-%m-%dT%H:%M:%S")
+    date_beg = datetime.fromisoformat(output_meta["DATE-BEG"].value)
+    date_end = datetime.fromisoformat(output_meta["DATE-END"].value)
+    date_avg = (date_beg + (date_end - date_beg) / 2).strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3]
     output_meta["DATE-AVG"] = date_avg
     output_meta["DATE-OBS"] = date_avg
```
svank: In other flows that run at fixed times with multiple inputs (e.g. the CAM flow), what we've done is set `DATE-AVG` to the mean of the `DATE-OBS` of all the input files, and set `DATE-OBS` to the nominal time. (E.g. CAMs always have `DATE-OBS` on the 32-minute grid, but `DATE-AVG` varies if some input files are missing).

It might be most robust to have the scheduler side pass `reference_time` to the core flow and use that for `DATE-OBS` here.

svank (changes requested):

punchbowl/auto/flows/velocity.py
svank: `from sqlalchemy.orm import aliased`
`from punchbowl.auto.control.db import FileRelationship`
```
    child = aliased(File)
    child_exists_subquery = (session.query(FileRelationship)
                             .join(child, FileRelationship.child == child.file_id)
                             .filter(FileRelationship.parent == File.file_id)
                             .filter(child.file_type == "VA")
                             .exists())
    all_ready_files = (session.query(File)
                   .filter(File.state.in_(["created", "progressed"]))
                   .filter(File.level == "3")
                   .filter(File.file_type == "PT")
                   .filter(File.observatory == "M")
                   .filter(~child_exists_subquery)
                   .order_by(File.date_obs.desc()).all())
```

This will query for PTMs that haven't yet been used to make a VAM.

punchbowl/level3/tests/test_velocity.py
```diff
@@ -114,7 +119,7 @@ def test_with_bad_data(tmpdir):
     write_ndcube_to_fits(cube, file_path)
 
     with pytest.raises(ValueError):
-        _ = track_velocity([str(file_path)])
+        _ = track_velocity([str(file_path)], reference_time=datetime.fromisoformat(meta["DATE-OBS"].value))
```
jmbhughes: Why assign it to an empty variable? Why not just run it?

jmbhughes (commented):
This is probably all good. I would like to carefully check it in our test before release though.

punchbowl/level3/velocity.py
```diff
@@ -515,9 +519,9 @@ def track_velocity(files: list[str],
     """
     # Set defaults for missing input parameters
     if ycens is None:
-        ycens = np.arange(7, 14.5, 0.5)
+        ycens = np.arange(30, 90, 10)
```
courtneypeck, line 522: ycens looks like a magic number. Add a comment to provide some context on why this array is used if no input is given.

courtneypeck (approved):

jmbhughes (approved):

PR #1019, opened 2026-06-30
QuickPUNCH should run from clears now instead of polarized

jmbhughes:
This is the last step for getting the new QuickPUNCH products out the door.

PR #1088, opened 2026-07-22
Add QuickPUNCH code to v0l-prod

jmbhughes:
It was approved in #1019 for main so this takes care of putting it in prod.


jmbhughes:
pre-commit.ci autofix

PR #1132, opened 2026-08-04
Implement daily quicklook movies and images

### LINKED ISSUE #757: Make daily quicklook movies
Problem: Daily quicklook movies to be provided in public domain for end users. We are currently generating these movies internally.
Appetite: Only a couple days
Solution:
- [ ] Make daily quicklook movies with total brightness generated using L3 polarized data, with proper intensity scaling and gamma.
- [ ] Q: where will they be disseminated through? SDAC?

Rabbit holes:
Datapath can be accessed by someone having server access.

No-gos:
Don't expect all the movies will be beautiful with the settings done.
Quicklook movies are not meant for scientific analysis.

### LINKED ISSUE #899: Make JPEG files (as opposed to JP2 only) each day as well
This is a request from Sarah.

lowderchris:
## PR summary

Implements daily quicklook movies and images

## Todos

- [x] Iterate forwards or backwards?
- [ ] Is there a way to avoid having to iterate at all?
- [x] Don't include current day?
- [x] Add docstrings

## Test plan

Will try in production

## AI usage

Optimization of some if-statement syntax.
Debugging some flow logic.


lowderchris:
Will need to add this to the config:
```
  quicklook:
    run-on: "phoenix"
    enabled: true
    description: "Creates quicklook movies and images."
    tags: [ "quicklook" ]
    priority:
      initial: 10
    seconds: [60]
    escalation: [1000]
    schedule: "59 23 * * *"
    launch_weight: 5.0
    start_time: "2025-06-21T00:00:00"
    max_wait_days: 30
    flows_max: 30
    expected_files:
      L3_CAM: 45
      L3_PAM: 45
      L3_CTM: 180
      L3_PTM: 360
      LQ_QAM: 45
      LQ_QNN: 180
```

Note that potentially for the first batch of re-processing, max_wait_days might need to be stretched to wait for things to catch up...

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+from punchbowl.auto.control.db import File, Flow, Quicklook
+from punchbowl.auto.control.util import get_database_session, load_pipeline_configuration, load_quicklook_scaling
+from punchbowl.auto.flows.util import file_name_to_full_path
+from punchbowl.data.punch_io import load_ndcube_from_fits, write_ndcube_to_quicklook
+from punchbowl.data.visualize import animate_punch
+from punchbowl.prefect import get_logger
+
+
+@task(cache_policy=NO_CACHE)
+def visualize_query_ready_files(session,
```
courtneypeck, line 19: I probably don't have to tell you what you need to add here 🚔

punchbowl/auto/control/db.py
```diff
@@ -80,6 +80,17 @@ def directory(self, root: str):
       File.observatory, File.file_version, File.date_obs,
       unique=True)
 
+
+class Quicklook(Base):
```
courtneypeck, line 84: It would be good to have a unit test that tests this database query.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+
+                make_image = (not image_made) and (len(files) > 0)
+                make_movie = (not movie_made) and (len(files) > movie_nfiles)
+
+                if make_image or make_movie:
+                    all_ready_files.append(list(files))
+                    all_product_codes.append(f"L{level}_{product_code}")
+                    all_tasks.append({"day": day, "level": level, "code": product_code,
+                                    "make_image": make_image, "make_movie": make_movie})
+
```
courtneypeck, line 69: Should we add a block that returns a warning if no movies or images were made?

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+
+    logger.info(f"{len(all_product_codes)} days will be visualized.")
+    return all_ready_files, all_product_codes, all_tasks
+
+
+@task(cache_policy=NO_CACHE)
+def visualize_flow_info(input_files: list[File],
```
courtneypeck, line 76: Docstrings please!

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+        state=state,
+        flow_level="M",
+        creation_time=creation_time,
+        priority=priority,
+        call_data=call_data,
+    )
+
+
+@flow
+def quicklook_scheduler_flow(pipeline_config_path=None,
```
courtneypeck, line 116: Docstrings please!

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+    for file_list, product_code, task in zip(file_lists, product_codes, tasks):
+        flow = visualize_flow_info(file_list, product_code, pipeline_config, task,
+                                   framerate=framerate, resolution=resolution)
+        session.add(flow)
+
+    session.commit()
+
+def generate_flow_run_name():
```
courtneypeck, line 139: Docstrings please!

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    parameters = flow_run.parameters
+    code = parameters["product_code"]
+    files = parameters["file_list"]
+    return f"movie-{code}-len={len(files)}-{datetime.now()}"
+
+
+@flow(flow_run_name=generate_flow_run_name)
+def quicklook_core_flow(file_list: list,
```
courtneypeck, line 147: Docstrings please!

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    os.makedirs(os.path.dirname(path_image), exist_ok=True)
+
+    if make_image:
+        write_ndcube_to_quicklook(cube, filename=path_image, vmin=vmin, vmax=vmax)
+    if make_movie:
+        animate_punch(file_list, output_path=path_movie, fps=framerate, n_jobs=12, vmin=vmin, vmax=vmax)
+
+
+@flow
+def quicklook_process_flow(flow_id: int, pipeline_config_path=None, session=None):
```
courtneypeck, line 167: Docstrings please!

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+@flow(flow_run_name=generate_flow_run_name)
+def quicklook_core_flow(file_list: list,
+                        output_movie_dir: str,
+                        make_image: bool,
+                        make_movie: bool,
+                        framerate: int = 10) -> None:
+    cube = load_ndcube_from_fits(file_list[0])
+    vmin, vmax = load_quicklook_scaling(level=cube.meta["LEVEL"].value, product=cube.meta["TYPECODE"].value, obscode=cube.meta["OBSCODE"].value)
```
courtneypeck, line 153: I don't understand why we keep the vmin and vmax values for quicklooks in the punchpipe_config.yaml (rather than in a more generic constants file) since this config file seems really specific to the flows. For example, it would be really useful for animate_punch to also determine default vmin and vmax values using this quicklook scaling. This would be out of scope for this ticket, but I'm curious to hear your opinions on whether we should change this in a future ticket.

courtneypeck (changes requested):
Needs lots of docstrings! 🚔 Otherwise, I think it's pretty close.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    all_product_codes = []
+    all_tasks = []
+
+    code_mapping = {"3": ["CA", "PA", "CT", "PT"],
+                    "Q": ["QA", "QN"]}
+
+    expected_files = pipeline_config["flows"]["quicklook"]["expected_files"]
+
+    day = datetime.fromisoformat(pipeline_config["flows"]["quicklook"]["start_time"])
+    reference_time = reference_time.replace(hour=0, minute=0, second=0, microsecond=0)
```
jmbhughes, line 34: I like how this is done. You've taught me something here. I didn't know there was a `replace` or `fromisoformat`.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+
+    while day <= reference_time:
```
jmbhughes, line 36: I think a high level comment about the approach could help make this more readable. Something to the effect that "we're stepping backwards through the images to collect..."

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+                                     .filter(Quicklook.level == level)
+                                     .filter(Quicklook.code == product_code).all())
+                image_made = quicklook_results.image_made if quicklook_results else False
+                movie_made = quicklook_results.movie_made if quicklook_results else False
+
+                if image_made and movie_made:
+                    continue
+
+                files = (session.query(File)
+                        .filter(File.state.in_(["created", "progressed", "quickpunched"]))
```
jmbhughes, line 51: I guess you're just avoiding `creating` or `planned` kinds of states?

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    call_data = json.dumps(
+        {
+            "file_list": [input_file.filename() for input_file in input_files],
+            "product_code": product_code,
+            "day": task_item["day"].isoformat(),
+            "level": task_item["level"],
+            "code": task_item["code"],
+            "make_image": task_item["make_image"],
+            "make_movie": task_item["make_movie"],
+            "output_movie_dir": os.path.join("movies", out_path),
```
jmbhughes, line 99: Is this the right out path?

jmbhughes: What's the root of it? Should it append something before the `movies` bit?

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+        flow_level="M",
+        creation_time=creation_time,
+        priority=priority,
+        call_data=call_data,
+    )
+
+
+@flow
+def quicklook_scheduler_flow(pipeline_config_path=None,
+                             session=None,
```
jmbhughes, line 117: I recently learned that you're not supposed to pass sqlalchemy sessions around. We should open a ticket to rectify this problem.

punchbowl/auto/flows/quicklook.py
jmbhughes: It's an idea worth considering. The only point I see against it is that a constants file might be harder to change. But that could also be a point in its favor because we don't want to confuse the community by changing these often.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,228 @@
...
+    flow_db_entry.start_time = datetime.now()
+    session.commit()
+
+    # load the call data and launch the core flow
+    flow_call_data = json.loads(flow_db_entry.call_data)
+
+    day = datetime.fromisoformat(flow_call_data.pop("day"))
+    level = flow_call_data.pop("level")
+    code = flow_call_data.pop("code")
+    make_image = flow_call_data.pop("make_image")
```
jmbhughes, line 191: I'm confused. Why pop these variables to put them back in a few lines later?

jmbhughes (changes requested):
This doesn't yet deal with the pile up of many runs possibly or the chance that we could get stuck trying to process the same thing over and over does it?

punchbowl/auto/flows/quicklook.py
courtneypeck: I made a ticket for this to extract this discussion from this PR: https://github.com/punch-mission/punchbowl/issues/1191

courtneypeck (approved):
Thanks for adding docstrings!

punchbowl/auto/flows/quicklook.py
lowderchris: Ticket opened.

punchbowl/auto/flows/quicklook.py
lowderchris: I think it made more sense when we were rapidly iterating and needed something adjustable in the field. I'm in favor of this.

punchbowl/auto/flows/quicklook.py
lowderchris: I wanted to make sure those files were existent before proceeding. Would it be smart enough to wait?

punchbowl/auto/flows/quicklook.py
lowderchris: Depending on the timing of running this, once it finishes the backlog of quicklook, it may often return empty task lists. It should just schedule nothing, but I'll add a logger note of that.

punchbowl/auto/flows/quicklook.py
lowderchris: Some legacy code from the original daily operator movies I didn't catch. Thanks! Should be fixed now.

punchbowl/auto/flows/quicklook.py
jmbhughes: You're probably absolutely right. Sam encountered a problem recently where `planned` states were getting selected. So good call.

punchbowl/auto/control/tests/test_db.py
```diff
@@ -0,0 +1,49 @@
...
+    result = db.query(Quicklook).filter_by(code="PAM").one()
+    assert result.movie_made is False
+    assert result.image_made is False
+
+
+def test_quicklook_insert(db_empty):
+    row = Quicklook(day=datetime(2026, 1, 2), level="3", code="CAM")
+    db_empty.add(row)
+    db_empty.commit()
+    assert db_empty.query(Quicklook).count() == 1
```
jmbhughes, line 49: This test seems a bit simplistic. I'm not sure what its purpose is. I'd like to see it modified so that it sets up a situation where the quicklook flow will create outputs and then confirms that a quicklook database entry is made after running the flow.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,314 @@
...
+        raise e
+    else:
+        flow_db_entry.state = "completed"
+        flow_db_entry.end_time = datetime.now()
+
+        quicklook_query = (session.query(Quicklook)
+                           .filter(Quicklook.day == day)
+                           .filter(Quicklook.level == level)
+                           .filter(Quicklook.code == code))
+        if quicklook_query is None:
```
jmbhughes: You created the `quicklook_query` so that prevents it from being `None` I think. Maybe you're meaning to call `.all()` at the end of that line to see if there was a query in existence?

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,314 @@
...
+            quicklook_query = Quicklook(day=day, level=level, code=code)
```
jmbhughes: Shouldn't the `Quicklook` database entry also have `nfiles`?

punchbowl/auto/flows/quicklook.py
jmbhughes: I think this little bit of logic and the `if` are probably not necessary.

punchbowl/auto/control/tests/test_db.py
jmbhughes: You might even parameterize it by the number of images so you could run two tests, one where the Quicklook flow is triggered and one where it's not to test both branches.

jmbhughes (changes requested):
I have a few comments but my main concern is whether this will trigger if we're missing a few images. It seems that it'll hang in that case and never make quicklook for that day (unless I'm missing something which is highly possible). We should probably add a max wait time so it'll go ahead and make quicklook even if there aren't sufficient files after a configurable wait time.

punchbowl/auto/flows/quicklook.py
jmbhughes: Oops. I didn't see that farther down.

punchbowl/auto/flows/quicklook.py
lowderchris: Agreed. Removing.

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,315 @@
...
+                        .filter(File.file_type == product_code[0:2])
+                        .filter(File.observatory == product_code[2])
+                        .order_by(File.date_obs.asc()).all())
+
+                movie_nfiles = expected_files[f"L{level}_{product_code}"]
+
+                make_image = (not image_made) and (len(files) > 0)
+                if (reference_time - day) < timedelta(days=max_wait_days):
+                    make_movie = (not movie_made) and (len(files) >= movie_nfiles)
+                else:
```
jmbhughes:
```suggestion
                else: # we've been waiting more than the configured wait time
```

jmbhughes (approved):

punchbowl/auto/flows/quicklook.py
```diff
@@ -0,0 +1,315 @@
...
+    all_product_codes = []
+    all_tasks = []
+
+    code_mapping = {"3": ["CAM", "PAM", "CTM", "PTM"],
+                    "Q": ["QAM", "QNN"]}
+
+    expected_files = pipeline_config["flows"]["quicklook"]["expected_files"]
+    max_wait_days = pipeline_config["flows"]["quicklook"]["max_wait_days"]
+
+    flows_max = pipeline_config["flows"]["quicklook"]["flows_days"]
```
jmbhughes: I think this should be "flows_max"

PR #1134, opened 2026-08-07
Flips starfield axis during subtraction

lowderchris:
## PR summary

A errant flip was occurring in one of the starfield model axes during the subtraction process, when determining a union WCS encompassing models. The WCS does need to be flipped along the RA axis to account for this factor of -1 with the cdelt. However, the code was making a reference to the WCS, which resulted in the flip being undone.

This corrects that by making a deepcopy of the model WCS to avoid flipping a flip.

## Todos

None

## Test plan

See below example plots testing with v0l data on punch190.

## Breaking changes

None

## Related issues

None

## AI usage

None.

### DIFF
--- a/changelog/1134.bugfix.rst
+++ b/changelog/1134.bugfix.rst
@@ -1,1 +1,1 @@
-Corrects a bug with starfield model cdelt during subtraction.
+Corrects a bug with starfield model subtraction where the RA-axis cdelt was being accidentally sign flipped twice.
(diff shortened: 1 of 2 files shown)


lowderchris:
Using the existing code on main:
sfield_subtract_before

and after this fix:
sfield_subtract_after

jmbhughes (approved):

PR #1135, opened 2026-08-11
#1001 put nfi back into mosiacs

### LINKED ISSUE #1001: Put NFI back into mosaics by removing todos and code disablement of NFI
After #1000, we're ready to deploy NFI. This requires hunting down all the spots NFI was disabled and combining it back into mosaics in the automation.

Rokkaan5:
## PR summary

*What does this PR introduce?*

This PR is supposed to undo all the commented out lines that were put in place to ignore NFI in the pipeline until the dynamic stray light could be dealt with. Now that we have the PCA method to initially release some visually pleasing NFI images, we want to incorporate those results into the mosaics.

## Todos

*Please list tasks to complete this PR.*

- [ ] Find and appropriately modify the todos that used to comment out NFI 
- [ ] "Test" that somehow nothing in the pipeline is critically broken from the changes made

## Test plan

*How does this PR verify the code works correctly?*

TBD...

## Breaking changes

*Document any breaking changes from this PR.*

## Related issues

*Link the issues related to this PR.*

This is the main task of ticket #1001 

## AI usage

*State how AI was used in this PR.*

### DIFF
--- a/punchbowl/auto/flows/level1.py
+++ b/punchbowl/auto/flows/level1.py
@@ -270,7 +270,7 @@ def get_psf_model_paths(level0_files, pipeline_config: dict, session=None):
               .order_by(File.file_version.desc(), File.date_obs.desc()).all())
     results = []
     for l0_file in level0_files:
-        
+
         target_type = PSF_MODEL_CORRESPONDING_TYPES[l0_file.file_type[1]]
         # We want to pick the latest model that's before the observation, so we go backwards in time, past any
         # later-in-time models, until we hit the first model that's before the observation.
@@ -290,7 +290,7 @@ def get_psf_model_paths(level0_files, pipeline_config: dict, session=None):
 
 def get_psf_model_path(level0_file, pipeline_config: dict, session=None, reference_time=None) -> str:
     psf_model_type = PSF_MODEL_CORRESPONDING_TYPES[level0_file.file_type[1]]
-    
+
     best_model = (session.query(File)
                   .filter(File.file_type == psf_model_type)
                   .filter(File.observatory == level0_file.observatory)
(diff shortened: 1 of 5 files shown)


Rokkaan5:
From what I understood, the action in commit 502186c, is supposed to be sufficient to get this task (of putting NFI back into PUNCH mosaics) for those specific "TODO" tags, but perhaps not sufficient to properly deal with the proper PSF processing for NFI. Ticket #1136 was created to make sure this note is accounted for.

Rokkaan5:
pre-commit.ci autofix

jmbhughes (commented):
Looks great. Can you make sure the CI passes?

punchbowl/auto/flows/level1.py
```diff
@@ -270,10 +270,7 @@ def get_psf_model_paths(level0_files, pipeline_config: dict, session=None):
               .order_by(File.file_version.desc(), File.date_obs.desc()).all())
     results = []
     for l0_file in level0_files:
-        # TODO - Turn this back on once fine tuned for NFI
-        if l0_file.observatory == "4":
-            results.append("")
-            continue
```
svank, line 276: For this and the two other PSF-related changes in this file, I wonder if we should avoid merging those changes until we have PSFs for NFI---otherwise test runs might be difficult. (I think the reason for this hack was because we couldn't get a good PSF for NFI, and we decided no PSF was better than a bad PSF.)

punchbowl/auto/flows/level2.py
```diff
@@ -34,8 +34,7 @@ def _level2_query_ready_files(session, polarized: bool, pipeline_config: dict, m
     logger = get_logger()
     all_ready_files = (session.query(File).filter(File.state == "created")
                        .filter(File.level == "1")
-                        # TODO: This line temporarily excludes NFI
-                       .filter(File.observatory.in_(["1", "2", "3"]))
+                       .filter(File.observatory.in_(["1", "2", "3", "4"]))
```
svank: This line can probably be removed altogether, since asking for these four observatories is just asking for everything. It doesn't *really* matter, except sometimes I've seen extra filters make DB queries slower if it makes the DB not use an index effectively.

svank (changes requested):

jmbhughes:
@Rokkaan5 @svank what is the plan with this PR given #1179 ?

PR #1141, opened 2026-08-19
Fix outlier flagging

### LINKED ISSUE #1087: Fix outlier flagging for L2+
A while back, we turned the OUTLIER flag into a bitmask for L2+, to indicate which of the input observatories had an outlier image. When we switched the F corona estimation to be per-WFI, we started writing per-WFI L2 X files which are the inputs to L3 (*not* the L2 CTM/PTM files). For a given trefoil/timestamp (i.e. one L2 PTM/CTM), it looks like if one of the input images is an outlier, *all* the L2 X files are being flagged as outliers, meaning that the L3 CIM/PIM always gets a bitmask of 14 = 01110 i.e. all WFIs are outliers (unless one of the WFIs is missing for that timestamp).

It looks like in the flow itself the X file OUTLIER flag is being set based on only the input images for that WFI, but on the scheduling side, the DB entry for all X files is being set to 1 if any of the input L1s are outliers, and that DB value gets propagated into the file before it's written out.

The L2 PTM/CTMs appear to have correct bitmask OUTLIER values.

s0larish:
## PR summary

- Fixed outlier flagging at L2+ data products
- Added a test to verify the flagging

## Related issues

Fix outlier flagging for L2+ #1087 

## AI usage

Debugging some errors in the test


s0larish:
pre-commit.ci autofix

punchbowl/auto/flows/level2.py
courtneypeck: Rename "obses" to "observatories"

courtneypeck, line 268: Oh dear. It looks like none of the functions in this file have proper docstrings. Could you at least add docstrings for this function since you have poked around in here recently?

courtneypeck (approved):

svank (approved):
Thanks Ritesh!

I think this will fix it, but we should remember to check the actual output files next time we do a test run.

s0larish:
pre-commit.ci autofix

jmbhughes (approved):

PR #1142, opened 2026-08-19
Fix checking for existing *AMs

svank:
## PR summary

This is supposed to be checking if this CAM already exists, but it's actually checking if a CTM exists at the CAM time (which it never will, since CAMs are every 32 minutes on the minute, and CTM timestamps have non-zero seconds). The situation is if some of the input CTMs don't get generated until well after others, we might make the CAM with the earlier batch of CTMs, and then when new ones show up the scheduler will want to make the CAM again, but we can't overwrite already-generated files. We usually avoid this problem by making incomplete groups of CTMs wait to be CAM-ified until a set amount of time after the last CTM was generated, but sometimes the delay for the last CTM can be longer than that wait period.

## Test plan

Next pipeline test run on 190. This specific bit of logic, as well as the "make incomplete groups of CTMs wait to be CAM-ified" logic could be candidates for unit tests.

## AI usage

None


jmbhughes:
pre-commit.ci autofix

jmbhughes (approved):
Glad we caught this.

PR #1145, opened 2026-08-25
Improves starfield model generation and subtraction

### LINKED ISSUE #1100: Investigate why the Milky Way survives starfield subtraction
The diffuse Milky Way brightness survives our starfield subtraction. Work in this ticket will compare before-and-after starfield subtracted WFI clear images to figure out why it is surviving and determine a path forward for fixing this issue in the pipeline.

lowderchris:
## PR summary

Improves starfield model generation and subtraction via:
- Removing median background subtraction that is no longer needed. This should improve capturing the Milky Way into the starfield models.
- Allowing for a single starfield model subtraction. This will allow a potential full-sky starfield model subtraction.

## Todos

- [x] make figures to illustrate

## Test plan

I've run this code in a notebook on punch190

## Breaking changes

None

## AI usage

None.

### DIFF
--- a/punchbowl/level3/stellar.py
+++ b/punchbowl/level3/stellar.py
@@ -346,8 +346,9 @@ def generate_starfield_background(
 
 @punch_task
 def subtract_starfield_background_task(data_object: PUNCHCube,
-                                       before_starfield_path: str | None,
-                                       after_starfield_path: str | None,
+                                       before_starfield_path: str | None = None,
+                                       after_starfield_path: str | None = None,
+                                       starfield_path: str | None = None,
                                        is_polarized: bool = False) -> PUNCHCube:
     """
     Subtracts a background starfield from an input data frame.
@@ -363,6 +364,8 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
         path to a PUNCHCube background starfield map centered before the observation
     after_starfield_path : str
         path to a PUNCHCube background starfield map centered after the observation
+    starfield_path : str
+        path to a single PUNCHCube background starfield map centered around the observation
     is_polarized : bool
         whether the data is polarized
 
@@ -375,12 +378,16 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
     logger = get_logger()
     logger.info("subtract_starfield_background started")
 
-    if before_starfield_path is None and after_starfield_path is None:
+    if before_starfield_path is None and after_starfield_path is None and starfield_path is None:
         output = data_object
         output.meta.history.add_now("LEVEL3-subtract_starfield_background",
                                            "starfield subtraction skipped since path is empty")
-    elif before_starfield_path is None or after_starfield_path is None:
+    elif (before_starfield_path is None or after_starfield_path is None) and starfield_path is None:
         raise InvalidDataError("subtract_starfield_background requires two input starfield models.")
+    elif starfield_path is not None:
+        star_datacube = load_ndcube_from_fits(starfield_path)
+        wcs_celestial = star_datacube.celestial_wcs
+        wcs_celestial.wcs.cdelt[0] = wcs_celestial.wcs.cdelt[0] * -1
     else:
         star_datacube_before = load_ndcube_from_fits(before_starfield_path)
         star_datacube_after = load_ndcube_from_fits(after_starfield_path)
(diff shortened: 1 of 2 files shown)


punchbowl/level3/stellar.py
```diff
@@ -379,12 +378,16 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
     logger = get_logger()
     logger.info("subtract_starfield_background started")
 
-    if before_starfield_path is None and after_starfield_path is None:
+    if before_starfield_path is None and after_starfield_path is None and starfield_path is None:
```
s0larish: This can be condensed something like:
if not any((before_starfield_path, after_starfield_path, starfield_path)):

s0larish (dismissed):

punchbowl/level3/stellar.py
```diff
@@ -452,42 +455,43 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
...
-        data_object.data[original_mask] = 0
-        data_object.data[~np.isfinite(data_object.data)] = 0
-
-        data_object.meta.history.add_now("LEVEL3-subtract_starfield_background", "subtracted starfield background")
-        output = polarize_celestial_to_solar(data_object) if is_polarized else data_object
+    original_mask = data_object.data == 0
+
+    # TODO - Think about where to do the interpolation at this stage...
+    # Is this going to require a change in the subtraction code to avoid more reprojections back and forth?
+    if is_polarized:
```
courtneypeck, line 465: It seems like the code is going to fail if this "if" from above is used since `star_datacube` won't be defined: `if before_starfield_path is None and after_starfield_path is None and starfield_path is None:`

Am I missing something? Should there be a return after that "if" statement?

punchbowl/level3/stellar.py
lowderchris: Ah, that's an elegant way to do that. Thanks! Incorporated.

punchbowl/level3/stellar.py
lowderchris: Ah, thank you for catching that! I've made some logic structure changes there to keep ruff happy, let me know if there's a more elegant / readable way to do that.

punchbowl/level3/stellar.py
```diff
@@ -379,12 +378,19 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
...
-    if before_starfield_path is None and after_starfield_path is None:
+    if not any((before_starfield_path, after_starfield_path, starfield_path)):
         output = data_object
         output.meta.history.add_now("LEVEL3-subtract_starfield_background",
                                            "starfield subtraction skipped since path is empty")
-    elif before_starfield_path is None or after_starfield_path is None:
+        return output
+
+    if (before_starfield_path is None or after_starfield_path is None) and starfield_path is None:
         raise InvalidDataError("subtract_starfield_background requires two input starfield models.")
```
svank:
```suggestion
        raise InvalidDataError("subtract_starfield_background requires at least one input starfield model.")
```

punchbowl/level3/stellar.py
```diff
@@ -452,42 +458,43 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
...
-            data_object.uncertainty.array[...] = np.sqrt(data_object.uncertainty.array**2 +
-                                                         subtracted.subtracted[1]**2)
-
-        # Reset the data to be zero in invalid regions
-        data_object.data[original_mask] = 0
-        data_object.data[~np.isfinite(data_object.data)] = 0
-
-        data_object.meta.history.add_now("LEVEL3-subtract_starfield_background", "subtracted starfield background")
-        output = polarize_celestial_to_solar(data_object) if is_polarized else data_object
+    original_mask = data_object.data == 0
```
svank: This hasn't to do with the changes in this PR, but we shouldn't flag a pixel as masked just because it's zero---that can happen naturally.

```suggestion
    original_mask = (data_object.data == 0) * ~np.isfinite(data_object.uncertainty)
```

punchbowl/level3/stellar.py
```diff
@@ -452,42 +458,43 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
...
+        data_object.data[...] = subtracted.subtracted[0]
+        data_object.uncertainty.array[...] = np.sqrt(data_object.uncertainty.array**2 +
+                                                        subtracted.subtracted[1]**2)
+
+    # Reset the data to be zero in invalid regions
+    data_object.data[original_mask] = 0
+    data_object.data[~np.isfinite(data_object.data)] = 0
+
+    data_object.meta.history.add_now("LEVEL3-subtract_starfield_background", "subtracted starfield background")
+    output = polarize_celestial_to_solar(data_object) if is_polarized else data_object
```
svank, line 496: Again, nothing to do with this PR, but should this be run on the starfield model, not the subtracted image?

lowderchris: We walked through some of the logic, but for posterity:
- PUNCHImageProcessor uses polarize_solar_to_celestial when used to load data.
- This is used when generating polarized starfield models, so input data are converted to that frame.
- When subtract_from_image is called via a starfield model, it uses the PUNCHImageProcessor to convert the input data into that celestial polarization frame.
- The subtracted output data is then converted back into the solar polarization frame.

This was good to double check the logic.

punchbowl/level3/stellar.py
```diff
@@ -452,42 +458,43 @@ def subtract_starfield_background_task(data_object: PUNCHCube,
...
-
-        # Reset the data to be zero in invalid regions
-        data_object.data[original_mask] = 0
-        data_object.data[~np.isfinite(data_object.data)] = 0
-
-        data_object.meta.history.add_now("LEVEL3-subtract_starfield_background", "subtracted starfield background")
-        output = polarize_celestial_to_solar(data_object) if is_polarized else data_object
+    original_mask = (data_object.data == 0) * ~np.isfinite(data_object.uncertainty)
+
+    # TODO - Think about where to do the interpolation at this stage...
```
jmbhughes: Is this TODO still active?

courtneypeck (approved):

punchbowl/level3/stellar.py
lowderchris: TODO no longer needed. Removing.

jmbhughes (approved):

PR #1152, opened 2026-08-27
Disable NFI static stray light processing

### LINKED ISSUE #1148: Fix stray light generation bug with NFI data
Stray light generation processes are failing for NFI data, where the obs-code of 4 is out of range for what has up to this point been expected for the three WFIs.

courtneypeck:
## PR summary

Disables NFI statis stray light processing in the construct_stray_light_scheduler_flow

## Todos

None

## Test plan

I will need input from someone on the team on how to test this change

## Breaking changes

Unclear

## Related issues

N/A

## AI usage

None

### DIFF
--- a/punchbowl/auto/flows/stray_light.py
+++ b/punchbowl/auto/flows/stray_light.py
@@ -256,7 +256,6 @@ def construct_stray_light_scheduler_flow(pipeline_config_path=None, session=None
     for i in range(n, -1, -1):
         t = t0 + i * increment
         for model_type in ["SR", "SM", "SZ", "SP"]:
-            #for observatory in ["1", "2", "3", "4"]:
             for observatory in ["1", "2", "3"]: # Disable NFI static stray light processing
                 key = (model_type, observatory, t)
                 model = existing_models.get(key)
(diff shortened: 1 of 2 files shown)


punchbowl/auto/flows/stray_light.py
```diff
@@ -256,7 +256,8 @@ def construct_stray_light_scheduler_flow(pipeline_config_path=None, session=None
     for i in range(n, -1, -1):
         t = t0 + i * increment
         for model_type in ["SR", "SM", "SZ", "SP"]:
-            for observatory in ["1", "2", "3", "4"]:
+            #for observatory in ["1", "2", "3", "4"]:
```
lowderchris:
```suggestion
```
Maybe kill this line and leave in the comment on the line below a note about not including 4 for NFI?

lowderchris (approved):

jmbhughes:
I'd rather fix it than just disable it. We need it.

courtneypeck:
> I'd rather fix it than just disable it. We need it.

Can you elaborate on this? Sam says that the stray light model isn't built for NFI and that it needed to be disabled (see Slack conversation from 8/26 on punch-notify.

jmbhughes:
> > I'd rather fix it than just disable it. We need it.
> 
> Can you elaborate on this? Sam says that the stray light model isn't built for NFI and that it needed to be disabled (see Slack conversation from 8/26 on punch-notify.

You should listen to Sam and ignore me haha. The point I was trying to make was that I wanted to see NFI working properly instead of just disabling the stray light. But you're both right that NFI PCA doesn't need stray light so we should disable it for now.

jmbhughes (approved):

PR #1159, opened 2026-09-04
Fix #936 by clipping to 1 instead of 0 for transparency

### LINKED ISSUE #936: Unclip JP2 products to zero at negative values
We need to better handle negative values. Right now, they're clipping to 0 which results in a transparent patch in Helioviewer applications.

jmbhughes:
We were previously clipping all negative values to 0. This led to negative values becoming transparent in JHelioviewer. This fixes that.


svank:
lgtm

PR #1184, opened 2026-09-15
New PCA for QuickPUNCH

svank:
## PR summary

Here's the current draft, rebased on v0l-prod

## Todos

While this is runnable, in the main PR we'll still want

- [ ] docs
- [ ] ruff satisfaction---type annotations, etc
- [ ] more tests?
- [ ] and the rest of the science pipeline, of course.

It seems like we're leaning towards skipping that for now to get this deployed on QuickPUNCH ASAP

## Test plan

Has run on 190

## Breaking changes

All of NFI processing is changed

## AI usage

None


svank:
The changes outside NFI processing are:

- `generic_process_flow` can compress and write files in parallel (because writing a batch of 2800 files serially was taking longer than the entirety of PCA filtering, reprojecting and post-processing). This is off by default because we normally don't need it
- `generic_process_flow` can be told to allow expected output files to be missing. Occasionally the DB says a file exists that doesn't exist on-disk. It would be nice if we could fix the cause, but for now the flow continues and of course doesn't write an output for that input.
- `generic_scheduler_flow_logic` can accept a function to create the file relationships, since this gets complicated for the PCA flow. (There's a one-to-one between filtered files and input files, but we also output an instrument-frame background and a PCA components file which have all the input files as parent files.)
- Small addition to the `group_files_by_time` function also used by other flows.
- Additions to F corona scheduling to support NFI and allow it to be configured separately
- Changes to omniheader and the `Level*.yaml` files

svank:
And caveat that this has not gotten a once-over to clean and comment things. I could do that first, it just comes down to how ASAP we want this deployed.

jmbhughes (approved):

PR #1188, opened 2026-09-17
Generate sha files from buffered FITS contents, rather than re-reading from disk

svank:
## PR summary

The PCA changes allow output files to be written in parallel, but even then in production it takes more than 2x as long to write the output files than to do the entirety of PCA filtering, reprojection and post-processing. This PR tries to speed up writing files.

Previously, we wrote FITS files to disk, and then read the contents back from disk to compute the sha hash. With this PR, we write the FITS file to an in-memory buffer, use that to compute the hash, and then write the buffer to disk. This avoids the round-trip of reading the FITS contents back, which I'm speculating incurs latency over NFS.

The benchmarking is noisy because I ran it on chimera while the pipeline was active, but I did 50 interleaved reps of "old method", then "new buffered method", then "unbuffered, only writing FITS, no hash file". "Buffered" is the clear winner, though I don't know why "no hash" is so much closer to "buffered" than "unbuffered"---maybe astropy writes FITS files in chunks, and it works better to send everything to griffin in one go? (I'm starting to think that's what's really happening here, and it's not much to do with the hash files at all.)

trial

This looks like a 2x speed up. The PCA output is taking 2+ hours, so if these results hold that's an hour right there!

The improvement vanishes when writing to a local SSD, so I think this is clearly an NFS latency thing.

## Todos

- [ ] Doesn't have to be here, but we can do similar for the JP2 code

## Test plan

I've verified that the hash files still match the FITS file, and I've verified that the FITS data array comes out identical after this change.

## AI usage

None


svank:
pre-commit.ci autofix

jmbhughes (approved):

svank:
Here's an updated benchmark that includes "buffered, no hash". It appears it really is mostly the "writing the fits file to a buffer" that's faster, not the "hashing from a buffer". ¯\\_(ツ)_/¯

trial

jmbhughes:
@svank it looks like there's a batch of files in your most recent plot that are equally fast for all methods (rep 10 area). What do you think causes that?

svank:
Dunno. If this is due to NFS latency, maybe that happened to be a period where few other things were reading/writing.

